PR #53, opened 2020-05-29
Actually track progress for send/receive

hannahhoward:
# Goals

Setup hooks with graphsync to facilitate tracking progress on the request -- this is important to get to being able to pause and resume at key increments. This sets up a bunch of the needed infrastructure for doing pause/resume and other needed facilities for retrieval.

# Implementation

- Modify the channels list to make send and receive mutable. Sync it's just two numbers, rather than all of the channels list to pointers to channel state, just make these members pointers to unsigned uint64 and use sync.atomic (low level primitives for doing atomic operations on numbers). Also, move the actual data structures for channels into the channels list package and leave the types as interfaces instead
- Setup an abstraction layer for converting between events that happen in graphsync through hooks and semantic events that mean something to data transfer. this allows some seperation of graphsync logic from data transfer logic. While I didn't attempt doing it here, I think we could move everything that is actually transport-layer specific behind and interface and have all the remaining high level logic be universal (and potentially use a different transport layer on different requests so we can support multiple transports in a single data transfer instance)
- Setup all hooks to actually record data progress on a data transfer so we can support pausing at intervals for retrieval


impl/graphsync/hooks/hooks_test.go
```diff
@@ -0,0 +1,377 @@
...
+				gsData.incomingRequestHook()
+			},
+			check: func(t *testing.T, events *fakeEvents, gsData *graphsyncTestData) {
+				require.True(t, events.OnRequestReceivedCalled)
+				require.False(t, gsData.incomingRequestHookActions.Validated)
+				require.Error(t, gsData.incomingRequestHookActions.TerminationError)
+			},
+		},
+		"recognized incoming request will record outgoing blocks": {
+			action: func(gsData *graphsyncTestData) {
```
shannonwells, line 121: You could reasonably combine this with the first valid incoming request test.

impl/graphsync/hooks/hooks_test.go
```diff
@@ -0,0 +1,377 @@
...
+		t.Run(testCase, func(t *testing.T) {
+			peers := testutil.GeneratePeers(2)
+			transferID := datatransfer.TransferID(rand.Uint64())
+			channelID := datatransfer.ChannelID{Initiator: peers[0], ID: transferID}
+			requestID := graphsync.RequestID(rand.Int31())
+			var request graphsync.RequestData
+			if data.makeRequest != nil {
+				request = data.makeRequest(requestID, channelID)
+			} else {
+				ext := &extension.TransferData{
```
shannonwells, line 210: Consider extracting these to a well-named function for readability.

PR #54, opened 2020-06-12
Single round trip pull

hannahhoward:
# Goals

This is basically the current state of improvements on data transfer

The only done feature is Single-round-trip pull requests -- rather than send a request/response on data transfer protocol for pull request we just immediately make a graphsync request and have the data transfer piggy back on it.

It also has a lot of the infrastructure for Intermedia vouchers and pause resume, but this is still a bit WIP. Still it's gotten large enough that it feels like it should have review now.

# Implementation

- Define interfaces for sending intermediate vouchers
- Add a voucher result to the response to a validation, to provide more information on why a voucher failed to validate or what other information may be neccesary to validate further (i.e. for intermediate vouchers)
- Add an update boolean to messages (for update messages as opposed to new channels)
- Add the ability to send intermediate voucher (note this does not do anything yet other than send it)
- Add pause boolean to the message (this will be used for pause unpause
- Switch to directly using messages as the content for graphsync extension
- Convert Data transfer pull flow to just make a single roundtrip over graphsync.


impl/graphsync/integration_test.go
```diff
@@ -0,0 +1,161 @@
...
+	voucher := testutil.FakeDTType{Data: "applesauce"}
+	sv := newSV()
+	sv.expectSuccessPull()
+	require.NoError(t, dt2.RegisterVoucherType(&testutil.FakeDTType{}, sv))
+
+	chid, err := dt1.OpenPushDataChannel(ctx, host2.ID(), &voucher, rootCid, gsData.AllSelector)
+	require.NoError(t, err)
+	opens := 0
+	completes := 0
+	sentIncrements := make([]uint64, 0, 21)
```
shannonwells, line 67: Where does 21 come from? I think it would be best in a named variable with a comment.

impl/graphsync/integration_test.go
```diff
@@ -0,0 +1,161 @@
...
+	gs1 := gsData.SetupGraphsyncHost1()
+	gs2 := gsData.SetupGraphsyncHost2()
+
+	dt1 := NewGraphSyncDataTransfer(host1, gs1, gsData.StoredCounter1)
+	dt2 := NewGraphSyncDataTransfer(host2, gs2, gsData.StoredCounter2)
+
+	finished := make(chan struct{}, 2)
+	errChan := make(chan struct{}, 2)
+	opened := make(chan struct{}, 2)
+	sent := make(chan uint64, 21)
```
shannonwells, line 110: see above

PR #57, opened 2020-07-24
Allow custom configuration of transports

hannahhoward:
# Goals

Support the end goal of loading/storing storage/retrieval deals from custom block stores

# Implementation

This is a bit fun/funky, but there are two basic adds:
1. A general purpose system for hooking in on a given voucher type to allow configuration of the the transport medium. (This is a bit of an abstraction breaker, cause the configurer receives a generic transport interface, that they cast to the transport medium they want to configure -- though I think this is ok, because once we have multiple transport options, the configurer may want to take different actions for different transports -- it feels silly to try to build a generic interface for configuring transports -- the things you do with each will be different)

Anyway this is done with by registering a transport configurer which has this signature:

```golang
type TransportConfigurer func(datatransfer.ChannelID, datatransfer.Voucher, datatransfer.transport)
```

(oy Configurer... that's a word -- maybe there is something better?)

2. A method for customizing the loader/storer for a given request on the graphsync transport medium. This uses graphsync's alternate persistence options under the hood.

You can see how this all comes together in the TestRoundTrip in integration test.

I also did some rearranging of files to avoid circular imports and break up the types file.


hannahhoward:
arg... I really didn't mean for this to seem like such a large PR -- the rearranging of files to avoid the circular imports really grew quick. I'd focus on graphsync transport changes, and the implementation of the registry of configuration functions

impl/integration_test.go
```diff
@@ -81,8 +117,54 @@ func TestRoundTrip(t *testing.T) {
...
+				})
+				require.NoError(t, err)
+			} else {
+				sourceDagService = gsData.DagService1
+			}
+			root, origBytes := testutil.LoadUnixFSFile(ctx, t, sourceDagService)
+			rootCid := root.(cidlink.Link).Cid
+
+			var destDagService ipldformat.DAGService
+			if data.customSourceStore {
```
ingar, line 145: should this be `data.customTargetStore`?

PR #71, opened 2020-09-02
Emit events with received cids

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/3417.

@hannahhoward This is step 1 which emits the receieved cids along with the channel event and state.

This PR is against a long running feature branch.


hannahhoward (approved):
LGTM. Be careful when you filter up to Storage/Retrieval that they may be relying on the old Progress event, so just keep an eye out there.

hannahhoward (changes requested):
Be careful when you filter up to Storage/Retrieval that they may be relying on the old Progress event, so just keep an eye out there.

Follow-up: I do not think we should record sentCIDs. We may be transferring pieces with 1000000s of CIDs and unless we have a direct use for them, putting all those on disk for every data transfer feels very expensive.

hannahhoward (approved):

hannahhoward:
Not blocking on changes cause it's just a feature branch and don't want to block you, but see feedback about sentCids -- there's a cost here in disk space if they aren't directly used.

aarshkshah1992:
@hannahhoward Not recording sent cids anymore.

PR #78, opened 2020-09-16
Restart Push integration tests and corresponding fixes

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/3417.

Depends on `go-graphsync` PR https://github.com/ipfs/go-graphsync/pull/93.

Restart test is in `restart_integration_test.go`


impl/events.go
```diff
@@ -152,6 +152,13 @@ func (m *manager) OnResponseReceived(chid datatransfer.ChannelID, response datat
 	return m.resumeOther(chid)
 }
 
+func (m *manager) OnRequestTimedOut(chid datatransfer.ChannelID) error {
+	// TODO Start a timer to cleanup state if this dosen't complete in a while
```
aarshkshah1992, line 156: @hannahhoward Have created  issue https://github.com/filecoin-project/lotus/issues/3878 to clean these up from the FSM.

channels/channel_state.go
```diff
@@ -15,6 +15,8 @@ import (
 
 // channelState is immutable channel data plus mutable state
 type channelState struct {
+	// peerId of the manager peer
```
hannahhoward, line 18: this is an interesting change and probably useful. 

I dunno if you saw but there's a method called `OtherParty` that takes a peer.ID for the party the data transfer is running on.

For symmetry, I'd like to do two things:
I'd like to rename ManagerPeer to ThisParty or SelfParty
and then I'd like to remove the parameter to OtherParty since we now have it in the channel state.
I guess I'd also be fine with SelfPeer and OtherPeer if Party doesn't make sense.

I'm ok with doing this as a seperate cleanup PR btw as well.

hannahhoward, line 18: this is an interesting change and probably useful. 

I dunno if you saw but there's a method called `OtherParty` that takes a peer.ID for the party the data transfer is running on.

For symmetry, I'd like to do two things:
I'd like to rename ManagerPeer to ThisParty or SelfParty
and then I'd like to remove the parameter to OtherParty since we now have it in the channel state.
I guess I'd also be fine with SelfPeer and OtherPeer if Party doesn't make sense.

I'm ok with doing this as a seperate cleanup PR btw as well.

go.mod
```diff
@@ -29,5 +29,6 @@ require (
 	github.com/libp2p/go-libp2p-core v0.5.0
 	github.com/stretchr/testify v1.5.1
 	github.com/whyrusleeping/cbor-gen v0.0.0-20200810223238-211df3b9e24c
+	go.uber.org/atomic v1.6.0
```
hannahhoward, line 32: did you mean to add this? I don't have an objection. I haven't worked with this library before but I'm sure I can figure it out.

impl/restart_integration_test.go
```diff
@@ -0,0 +1,291 @@
...
+			<-disConnChan
+			t.Logf("peers unlinked and disconnected, total increments received till now: %d", receivedTillNow.Load())
+
+			// Define function to wait for data transfer to complete
+			waitF := func(wait time.Duration, nCompletes int) (sentI, receivedI []uint64, err error) {
+				completes := 0
+
+				waitCtx, cancel := context.WithTimeout(rh.testCtx, wait)
+				defer cancel()
+				i := 0
```
hannahhoward, line 156: tad silly but is this variable actually used?

transport/graphsync/graphsync.go
```diff
@@ -84,7 +84,12 @@ func (t *Transport) OpenChannel(ctx context.Context,
 		return err
 	}
 	internalCtx, internalCancel := context.WithCancel(ctx)
+
 	t.dataLock.Lock()
+	// if we have an existing request pending for the channelID, cancel it first.
+	if cancelF, ok := t.contextCancelMap[channelID]; ok {
```
hannahhoward, line 90: non-blocking, something to consider: we may need to add a go routine that monitors transfers and cancels context for inactivity. reason being setting an overall timeout in go-fil-markets may not be a good sign of failure -- a large piece may just take a very long time. The real measure is transfer inactivity. anyway, let's cross that bridge later.

hannahhoward (approved):
All my comments are non-blocking except the naming of ManagerPeer, and we can put that in a seperate PR.

It looks like you're not actually disassembling the old data transfer instance in tests -- I guess that makes sense, since we don't really care what happens to it.

impl/restart_integration_test.go
aarshkshah1992: Great catch, not used anywhere. Have removed it.

go.mod
aarshkshah1992: Yeah, it's a simple wrapper over "sync/atomic.

channels/channel_state.go
aarshkshah1992: This is done.

transport/graphsync/graphsync.go
aarshkshah1992: I've added this note to https://github.com/filecoin-project/lotus/issues/3878, an issue I've created to track cleanups.

aarshkshah1992:
@hannahhoward Addressed review in https://github.com/filecoin-project/go-data-transfer/pull/79

PR #81, opened 2020-09-23
Complete channels that were cleaning up when we restart

aarshkshah1992:
For filecoin-project/lotus#3417.

### DIFF
--- a/channels/channels.go
+++ b/channels/channels.go
@@ -168,6 +168,10 @@ func (c *Channels) Restart(chid datatransfer.ChannelID) error {
 	return c.send(chid, datatransfer.Restart)
 }
 
+func (c *Channels) CompleteCleanupOnRestart(chid datatransfer.ChannelID) error {
+	return c.send(chid, datatransfer.CompleteCleanupOnRestart)
+}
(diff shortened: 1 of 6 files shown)


hannahhoward (changes requested):
We should probably either write an integration test or at minimum a unit test for RestartDataTransferChannel that verifies this works.

hannahhoward:
(if you want to merge as is and address comment seperately let me know and I'll remove the request changes)

hannahhoward:
I guess in fairness and integration test would be hard so maybe just a unit test

aarshkshah1992:
@hannahhoward I am not sure how to write a unit test for thus considering that as soon as a channel transitions to one of the penultimate states(Completing, Cancelling etc), the `CleanupChannel` handler for those states takes the channel to completion in an async way.

How do I get a channel that remains stuck rather than completing on those states so we can complete them on restart (for a NON-FLAKY test) ?

hannahhoward:
Fair enough, let's not worry bout it.

hannahhoward (approved):

PR #87, opened 2020-09-28
Disambiguate whether a revalidator recognized a request when checking for a need to revalidate

hannahhoward:
# Goals

Previously, the revalidators were very ambiguous in identifying if they had actually handled a
request for a revalidation check or not -- a "nil, nil" or "voucher result, nil" was interpreted as
not being handled. This is unfortunate, cause particularly a "voucher result, nil" is an indication
the request is handled by this check, and we should stop processing other revalidators that might
override the value with "nil, nil". While we could technically stop on a "voucher result, nil" -- it seems best to just make explicit whether we did or did not handle the request. 

# Implementation

Add a boolean to disambiguate whether the revalidator "recognized" a request for `OnPushDateReceived`, `OnPullDataSent`, and `OnComplete`

### DIFF
--- a/impl/events.go
+++ b/impl/events.go
@@ -32,10 +32,14 @@ func (m *manager) OnDataReceived(chid datatransfer.ChannelID, link ipld.Link, si
 	if chid.Initiator != m.peerID {
 		var result datatransfer.VoucherResult
 		var err error
+		var handled bool
 		_ = m.revalidators.Each(func(_ datatransfer.TypeIdentifier, _ encoding.Decoder, processor registry.Processor) error {
 			revalidator := processor.(datatransfer.Revalidator)
-			result, err = revalidator.OnPushDataReceived(chid, size)
-			return err
+			handled, result, err = revalidator.OnPushDataReceived(chid, size)
+			if handled {
+				return errors.New("stop processing")
+			}
+			return nil
 		})
 		if err != nil || result != nil {
 			msg, err := m.processRevalidationResult(chid, result, err)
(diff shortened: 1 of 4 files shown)


manager.go
```diff
@@ -31,20 +31,30 @@ type Revalidator interface {
 	// Revalidate revalidates a request with a new voucher
 	Revalidate(channelID ChannelID, voucher Voucher) (VoucherResult, error)
 	// OnPullDataSent is called on the responder side when more bytes are sent
-	// for a given pull request. It should return a VoucherResult + ErrPause to
+	// for a given pull request. The first value indicates whether the request was
+	// recognized by this revalidator and should be considered 'handled'. If true,
+	// the remaining to values are interpreted. If 'false' the request is passed on
```
aarshkshah1992, line 36: remaining  two*

impl/events.go
```diff
@@ -32,10 +32,14 @@ func (m *manager) OnDataReceived(chid datatransfer.ChannelID, link ipld.Link, si
...
 		var result datatransfer.VoucherResult
 		var err error
+		var handled bool
 		_ = m.revalidators.Each(func(_ datatransfer.TypeIdentifier, _ encoding.Decoder, processor registry.Processor) error {
 			revalidator := processor.(datatransfer.Revalidator)
-			result, err = revalidator.OnPushDataReceived(chid, size)
-			return err
+			handled, result, err = revalidator.OnPushDataReceived(chid, size)
+			if handled {
+				return errors.New("stop processing")
```
aarshkshah1992, line 40: Why do we need to return this error ? So that we break the `m.revalidators.Each` loop ?

Is there any other significance of this error ?

manager.go
aarshkshah1992: This typo needs to be fixed below as well.

aarshkshah1992 (approved):
LGTM except typos and one question.

impl/events.go
hannahhoward: yea that's what it does -- it stops the loop -- note that it's ignored in the return

PR #106, opened 2020-10-21
Correct data transfer Sent stats

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/4381.

@hannahhoward 

I still need into why some tests are failing. They depend on the `Sent` stats on the Channel and we have now changed how we tracked them i.e. we update them when we actually send the blocks. 

Let me know if you have any thoughts on those.


hannahhoward (commented):
This generally looks good -- I'm not sure why Pause and Resume are not working

aarshkshah1992:
@hannahhoward LGTM !

PR #112, opened 2020-10-31
Fix shutdown

hannahhoward:
# Goals

There were a number of problems in how we shutdown, which this aims to fix.
1. We were closing data transfer channels, which may prevent restart (unclear if changes were actually written to data store in practice)
2. We were failing to unregister from graphsync, which means if graphsync shutdown, we might face issues with channels appearing to complete with failure.

These issues were discovered while trying to improve the restart tests. Previously, the restart tests were not REALLY creating a proper restart -- the just booted up a new instance of graphsync and data transfer without shutting down the previous ones. When I tried to shutdown the previous ones, I found the errors listed above.

# Implementation

- improve libp2p test fixture by shutting down previous graphsync if you attempt to setup a new graphsync instance (graphsync shutdown happens by just cancelling the original graphsync context)
- actually shut down data transfer in restart tests
- shorten wait times in restart data transfer tests -- there was no reason for this and it mean tests went very long
- add a new shutdown method to the transport
- in graphsync transport shutdown, unregister all hooks and cancel all requests
- in data transfer stop method, remove close of data transfer channels
- update/improve mocks for transport + graphsync

### DIFF
--- a/impl/restart_integration_test.go
+++ b/impl/restart_integration_test.go
@@ -46,6 +46,8 @@ func TestRestartPush(t *testing.T) {
 			},
 			restartF: func(rh *restartHarness, chId datatransfer.ChannelID, subscriber datatransfer.Subscriber) {
 				var err error
+				require.NoError(t, rh.dt1.Stop(rh.testCtx))
+				time.Sleep(100 * time.Millisecond)
 				tp1 := rh.gsData.SetupGSTransportHost1()
 				rh.dt1, err = NewDataTransfer(rh.gsData.DtDs1, rh.gsData.DtNet1, tp1, rh.gsData.StoredCounter1)
 				require.NoError(rh.t, err)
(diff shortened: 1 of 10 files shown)


impl/restart_integration_test.go
```diff
@@ -173,15 +176,15 @@ func TestRestartPush(t *testing.T) {
 			}
 
 			// WAIT FOR TRANSFER TO COMPLETE -> THIS SHOULD NOT HAPPEN
-			sentI, receivedI, err := waitF(10*time.Second, 1)
+			sentI, receivedI, err := waitF(2*time.Second, 1)
 			require.EqualError(t, err, "context timed-out without completing data transfer")
 			require.True(t, len(receivedI) < totalIncrements)
 			require.NotEmpty(t, sentI)
+			t.Logf("not request was completed after disconnect")
```
ingar, line 183: Should this be "transfer was not completed after disconnect"?  Didn't quite understand the log

testutil/gstestdata.go
```diff
@@ -163,7 +165,12 @@ func NewGraphsyncTestingData(ctx context.Context, t *testing.T, host1Protocols [
 // SetupGraphsyncHost1 sets up a new, real graphsync instance on top of the first host
 func (gsData *GraphsyncTestingData) SetupGraphsyncHost1() graphsync.GraphExchange {
 	// setup graphsync
-	return gsimpl.New(gsData.Ctx, gsData.GsNet1, gsData.Loader1, gsData.Storer1)
+	if gsData.gs1Cancel != nil {
+		gsData.gs1Cancel()
+	}
```
ingar, line 170: Is this in case the test data is reused in a test?

ingar (approved):
Looks good, seems non-obvious.

testutil/gstestdata.go
hannahhoward: yes

PR #113, opened 2020-11-07
Reset timeouts

hannahhoward:
# Goals

Prevent Disconnects and Timeouts from shutting down the channel automatically if it doesn't close within an hour.

# Implementation

- Setup a series of channels that indicate the request has resumed operating (via receiving new data)
- Stop the go routine waiting for the timeout when the request resumes.

- couple other mutex things:
   - want to avoid a write lock in DataSent/DataReceived, so they use an RLock and then close the channel, checking to see if its already closed
   - then the actual channel is thrown out when we get another disconnect
   - not worried about race condition on multiple close because within a single request, calls to DataSent/DataReceived are sequential

### DIFF
--- a/impl/events.go
+++ b/impl/events.go
@@ -33,6 +33,20 @@ func (m *manager) OnDataReceived(chid datatransfer.ChannelID, link ipld.Link, si
 	if err != nil {
 		return err
 	}
+
+	m.reconnectsLk.RLock()
+	reconnect, ok := m.reconnects[chid]
+	var alreadyReconnected bool
+	select {
+	case <-reconnect:
+		alreadyReconnected = true
+	default:
+	}
+	if ok && !alreadyReconnected {
+		close(reconnect)
+	}
+	m.reconnectsLk.RUnlock()
+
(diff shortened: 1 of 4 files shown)


impl/events.go
```diff
@@ -169,10 +195,26 @@ func (m *manager) OnResponseReceived(chid datatransfer.ChannelID, response datat
...
+	if !ok || alreadyReconnected {
+		reconnect = make(chan struct{})
+		m.reconnects[chid] = reconnect
+	}
+	m.reconnectsLk.Unlock()
+	timer := time.NewTimer(m.channelRemoveTimeout)
+
 	go func() {
 		select {
 		case <-ctx.Done():
```
ingar, line 215: Do we need to do the channel cleanup in this case?

ingar (approved):
Just one question about the normal context cancellation case.

impl/events.go
hannahhoward: We have to cancel the context when a channel restarts, so this is important to do.

PR #115, opened 2020-11-11
Fix UseStore for restarts

hannahhoward:
# Goals

Fix UseStore in the case when a request is restarted -- sometimes this produces an error that causes a failure to use a custom store.

# Implementation

Unregister any existing stores when restarting.


dirkmc (approved):
LGTM 👍

PR #118, opened 2020-11-14
Allow data transfers with more than 8192 received CIDs

hannahhoward:
# Goals

Fix https://github.com/filecoin-project/lotus/issues/4856

# Implementation

This is a first attempt to fix 4856 -- it solves the issue in a protocol non-blocking way. There are more ideal ways to solve the issue but they involve changes to graphsync and require changes on the network.

What this does is remove the ReceivedCids list from the stored state for the channel, and it puts them instead in an append only file.

This has the effect of not only removing the limit but significantly reducing io & memory consumption relating to loading and saving these CIDs. The list is simply left in the file except when its needed for a restart.

The files are placed in a directory, indexed by channel ID.


cidlists/cidlists.go
```diff
@@ -0,0 +1,102 @@
...
+	if !info.IsDir() {
+		return nil, fmt.Errorf("%s is not a directory", base)
+	}
+	return &cidLists{
+		baseDir: base,
+	}, nil
+}
+
+// CreateList initializes a new CID list with the given initial cids (or can be empty) for a data transfer channel
+func (cl *cidLists) CreateList(chid datatransfer.ChannelID, initalCids []cid.Cid) error {
```
dirkmc, line 43:
```suggestion
func (cl *cidLists) CreateList(chid datatransfer.ChannelID, initialCids []cid.Cid) error {
```

cidlists/cidlists.go
```diff
@@ -0,0 +1,102 @@
...
+			if err == io.EOF {
+				return receivedCids, nil
+			}
+			return nil, err
+		}
+		receivedCids = append(receivedCids, c)
+	}
+}
+
+// DeleteList deletes the lsit for the given data transfer channel
```
dirkmc, line 94:
```suggestion
// DeleteList deletes the list for the given data transfer channel
```

cidlists/cidlists_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/stretchr/testify/require"
+
+	datatransfer "github.com/filecoin-project/go-data-transfer"
+	"github.com/filecoin-project/go-data-transfer/cidlists"
+	"github.com/filecoin-project/go-data-transfer/testutil"
+)
+
+func TestCIDLists(t *testing.T) {
+
+	err := os.MkdirAll("_test", 0755)
```
dirkmc, line 20: Maybe create _test in a tmp directory, then we don't need the .gitignore?

channels/channel_state.go
```diff
@@ -97,7 +96,8 @@ func (c channelState) Voucher() datatransfer.Voucher {
 
 // ReceivedCids returns the cids received so far on this channel
 func (c channelState) ReceivedCids() []cid.Cid {
-	return c.receivedCids
+	receivedCids, _ := c.channelCIDsReader(c.ChannelID())
```
dirkmc, line 99: Why not return the error from ReceivedCids?

cidlists/cidlists.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// CreateList initializes a new CID list with the given initial cids (or can be empty) for a data transfer channel
+func (cl *cidLists) CreateList(chid datatransfer.ChannelID, initalCids []cid.Cid) error {
+	f, err := os.Create(transferFilename(cl.baseDir, chid))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = f.Close()
```
dirkmc, line 49: May be worth returning the error on Close. It seems like the caller would want to know if the file couldn't be closed.

channels/channel_state.go
hannahhoward: just cause it's an API change -- and there's an interface that expects no error. But I will log it.

channels/channel_state.go
```diff
@@ -97,7 +96,11 @@ func (c channelState) Voucher() datatransfer.Voucher {
 
 // ReceivedCids returns the cids received so far on this channel
 func (c channelState) ReceivedCids() []cid.Cid {
-	return c.receivedCids
+	receivedCids, err := c.channelCIDsReader(c.ChannelID())
+	if err != nil {
+		log.Error(err)
```
Stebalien, line 101: I'd consider propagating this up and letting the caller handle it, but logging it and moving on is also quite reasonable as we have "no information" in this case.

cidlists/cidlists.go
```diff
@@ -0,0 +1,111 @@
...
+		err := cbg.WriteCid(f, c)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AppendList appends a single CID to the list for a given data transfer channel
+func (cl *cidLists) AppendList(chid datatransfer.ChannelID, c cid.Cid) (err error) {
```
Stebalien, line 64: Isn't this going to be really slow if we have to do this per cid? I'd suggest having some form of batching.

Stebalien (commented):
Bit of a late review: I'm concerned about the perf of opening a file per CID fetched.

An alternative approach would be to just write to the datastore. That is, write `/transferprefix/TRANSFER_ID/cids/CID`. We can then extract/delete the list with a range query. However, I'm also not entirely sure about the performance here either.

cidlists/cidlists.go
magik6k: Most files will be big, and the default lotus chunking is 1MB, so this should be usable

But it's still pretty far from ideal

PR #123, opened 2020-12-02
Wrap correct FSM error when closing a channel 

ingar:
## Summary
Fixing a typo, which caused the wrong error to be wrapped, preventing proper error inspection by callers


dirkmc (approved):
Nice catch 👍

PR #124, opened 2020-12-04
Better retry config

dirkmc:
Adds a `backoffFactor` option for network open stream retry configuration.

Previously the `backoffFactor` was set to the same value as `maxStreamOpenAttempts`.

Also fixes a bug where in practice one extra attempt would be made over the configured number of attempts.

### DIFF
--- a/network/libp2p_impl.go
+++ b/network/libp2p_impl.go
@@ -23,9 +23,17 @@ var log = logging.Logger("data_transfer_network")
 
 var sendMessageTimeout = time.Minute * 10
 
+// The max number of attempts to open a stream
 const defaultMaxStreamOpenAttempts = 5
+
+// The min backoff time between retries
 const defaultMinAttemptDuration = 1 * time.Second
+
+// The max backoff time between retries
 const defaultMaxAttemptDuration = 5 * time.Minute
+
+// The multiplier in the backoff time for each retry
+const defaultBackoffFactor = 5
 
 var defaultDataTransferProtocols = []protocol.ID{datatransfer.ProtocolDataTransfer1_1, datatransfer.ProtocolDataTransfer1_0}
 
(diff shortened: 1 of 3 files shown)


network/libp2p_impl.go
```diff
@@ -95,12 +106,16 @@ func (impl *libp2pDataTransferNetwork) openStream(ctx context.Context, id peer.I
 			return s, err
 		}
 
-		nAttempts := b.Attempt()
-		if nAttempts == impl.maxStreamOpenAttempts {
-			return nil, xerrors.Errorf("exhausted %d attempts but failed to open stream, err: %w", impl.maxStreamOpenAttempts, err)
+		// b.Attempt() starts from zero
+		nAttempts := b.Attempt() + 1
+		if nAttempts >= impl.maxStreamOpenAttempts {
```
nonsense, line 111: Not really related to this PR, but I am not sure why these counters are `float64`s (both `maxStreamOpenAttempts` and `b.Attempt()`?

network/libp2p_impl_test.go
```diff
@@ -174,3 +179,122 @@ func TestMessageSendAndReceive(t *testing.T) {
...
+	default:
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return w.Host.NewStream(ctx, p, pids...)
+}
+
+// TestSendMessageRetry verifies that the if the number of retry attempts
```
nonsense, line 202:
```suggestion
// TestSendMessageRetry verifies that if the number of retry attempts
```

nonsense (approved):
lgtm

network/libp2p_impl.go
dirkmc: Good question, I was wondering the same thing - basically it's because of the library we're using for the restart measurement

PR #127, opened 2020-12-11
Automatically restart push channel

dirkmc:
Implementation of auto-restart behaviour described in https://github.com/filecoin-project/go-fil-markets/pull/463#issuecomment-742576327

If a "push" data transfer channel stalls while transferring data, attempt to reconnect to the other party and send a "restart" request for the channel.

Note that the backoff behaviour on dial already exists [in the network layer](https://github.com/filecoin-project/go-data-transfer/blob/master/network/libp2p_impl.go#L109).

This PR adds a `pushChannelMonitor` to data-transfer `manager`. Each time the data-transfer `manager` opens a "push" data transfer channel, it adds the channel ID to the monitor.

Graphsync queues up data to be sent. The `pushChannelMonitor` periodically checks the amount of data queued against the amount sent. If the amount of pending data (queued - sent), is greater than the configured minimum amount over the configured interval (eg 1MB over 1s), the `pushChannelMonitor` assumes the transfer has stalled and attempts to send a "restart" request.

### DIFF
--- a/impl/integration_test.go
+++ b/impl/integration_test.go
@@ -513,7 +513,7 @@ func TestPushRequestAutoRestart(t *testing.T) {
 	tp1 := gsData.SetupGSTransportHost1()
 	tp2 := gsData.SetupGSTransportHost2()
 
-	restartConf := PushChannelRestartConfig(100*time.Millisecond, 1, 200*time.Millisecond)
+	restartConf := PushChannelRestartConfig(100*time.Millisecond, 1, 10, 200*time.Millisecond)
 	dt1, err := NewDataTransfer(gsData.DtDs1, gsData.TempDir1, gsData.DtNet1, tp1, gsData.StoredCounter1, restartConf)
 	require.NoError(t, err)
 	testutil.StartAndWaitForReady(ctx, t, dt1)
--- a/impl/integration_test.go
+++ b/impl/integration_test.go
@@ -209,7 +209,7 @@ func TestRoundTrip(t *testing.T) {
 				for opens < 2 || completes < 2 || len(sentIncrements) < 21 || len(receivedIncrements) < 21 {
 					select {
 					case <-ctx.Done():
-						t.Fatal("Did not complete succcessful data transfer")
+						t.Fatal("Did not complete successful data transfer")
 					case <-finished:
 						completes++
 					case <-opened:
@@ -343,7 +343,7 @@ func TestMultipleRoundTripMultipleStores(t *testing.T) {
 			for opens < 2*data.requestCount || completes < 2*data.requestCount {
 				select {
 				case <-ctx.Done():
-					t.Fatal("Did not complete succcessful data transfer")
+					t.Fatal("Did not complete successful data transfer")
 				case <-finished:
 					completes++
 				case <-opened:
@@ -481,7 +481,7 @@ func TestManyReceiversAtOnce(t *testing.T) {
 			for opens < 2*data.receiverCount || completes < 2*data.receiverCount {
 				select {
 				case <-ctx.Done():
-					t.Fatal("Did not complete succcessful data transfer")
+					t.Fatal("Did not complete successful data transfer")
 				case <-finished:
 					completes++
 				case <-opened:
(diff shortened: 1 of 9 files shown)


dirkmc:
The tests are actually all passing, there seems to be something out of sync between github and CI:
https://app.circleci.com/pipelines/github/filecoin-project/go-data-transfer?branch=feat%2Fpush-auto-retry
Screen Shot 2020-12-14 at 3 16 46 PM

impl/pushchannelmonitor.go
```diff
@@ -0,0 +1,254 @@
...
+func (m *pushChannelMonitor) start() {
+	go m.run()
+}
+
+func (m *pushChannelMonitor) run() {
+	defer m.onShutdown()
+
+	// Check the data-rate of all monitored channels on every tick
+	var tickerChan <-chan time.Time
+	if m.enabled() {
```
nonsense, line 98: Shouldn't we exit `run()` if `m.enabled()` is `false`?

impl/integration_test.go
```diff
@@ -497,6 +497,101 @@ func TestManyReceiversAtOnce(t *testing.T) {
...
+	require.NotNil(t, conn)
+
+	t.Logf("Waiting for auto-restart on push channel %s", chid)
+
+	(func() {
+		finishedCount := 0
+		for {
+			select {
+			case <-ctx.Done():
+				t.Fatal("Did not complete succcessful data transfer")
```
nonsense, line 579:
```suggestion
				t.Fatal("Did not complete successful data transfer")
```

nonsense (approved):
lgtm

impl/pushchannelmonitor.go
hannahhoward: agree on the "enabled" design being a little wonky -- presumably if any point someone calls setConfig(), then "enabled" becomes true, right?

but based on the code as written, it's not actually enabled, cause the Ticker never tics.

I feel like:
- move towards clarity that enabled actually can't be changed (would help if this were it's own package, and enabled was a private function where Start & SetConfig were not)
- allow calling setconfig to actually enable the monitor even is start was called earlier

hannahhoward (commented):
Something you should be aware of:
Graphsync has it's own notion of backpressure -- it will stop queuing when a certain among of memory has been allocated but not sent. We built this to resolve graphsync hogging memory if the network slowed down. So, that means, for this to work as a detection mechanism, the value of the diff between queued & sent must be lower than the backpressure default in graphsync. (I believe 16MB per peer). There aren't any defaults listed here, but I think you should inspect the Graphsync defaults: https://github.com/ipfs/go-graphsync/blob/master/impl/graphsync.go#L32, and your defaults which I assume are in the PR on markets, plus make sure Lotus isn't overriding the graphsync defaults (which would not surprise me)

hannahhoward:
Follow-up: One thing just to factor for the future is that while retrievals seem to be almost no one's priority, we're gonna have to deal with this problem eventually for Pulls too. And, while the https://github.com/ipfs/go-graphsync/issues/129 might help, we may also want to do some kind of keep alive in go-graphsync? I guess maybe go-yamux does that already.

dirkmc:
@hannahhoward for pulls I'm not sure we need this mechanism, I think it may make more sense to rely on go-graphsync itself to do retries. go-graphsync has some retry capability but it would be nice if it worked the same way as the retries in go-fil-markets and go-data-transfer

hannahhoward (approved):
LGTM -- just keep an eye out for the back pressure issue when you get to LOTUS

pushchannelmonitor/pushchannelmonitor.go
```diff
@@ -1,57 +1,59 @@
...
 	mgr  monitorAPI
-	cfg  *pushMonitorConfig
+	cfg  *PushMonitorConfig
 
 	lk       sync.RWMutex
 	channels map[*monitoredChannel]struct{}
 }
 
-type pushMonitorConfig struct {
+type PushMonitorConfig struct {
```
nonsense: Nitpick, but the package is called `pushchannelmonitor`, so these could be truncated to just `Monitor` and `Config`. Up to you, but I think this would help a bit with readability :)

dirkmc:
I realized that the data-rate monitoring mechanism I implemented was not very granular. If for example the pending data suddenly increased right before checking the data rate, it would appear that the data rate was too low and the monitor would trigger a channel restart.

The last commit instead adds functionality such that:
- the data-rate is checked and recorded multiple (configurable) times per interval, eg 10
- each check compares the sent amount between the start and end of the interval against the minimum required data rate

nonsense:
The new algorithm looks correct to me.

nonsense (approved):

PR #129, opened 2020-12-17
limit consecutive restarts with no data transfer

dirkmc:
Needed for https://github.com/filecoin-project/lotus/issues/5211

Currently when a push channel stalls (no data is sent for a while) the push channel monitor will attempt to restart the channel.
It's possible that the monitor will keep restarting the channel forever, and the transfer will get stuck.

This PR adds a configurable parameter `maxConsecutiveRestarts` - the maximum number of restarts in a row to attempt where no data is transferred. When the limit is reached the channel is closed.

### DIFF
--- a/pushchannelmonitor/pushchannelmonitor_test.go
+++ b/pushchannelmonitor/pushchannelmonitor_test.go
@@ -58,12 +58,14 @@ func TestPushChannelMonitorAutoRestart(t *testing.T) {
 			mockAPI := newMockMonitorAPI(ch, tc.errOnRestart)
 
 			m := NewMonitor(mockAPI, &Config{
-				Interval:          10 * time.Millisecond,
-				ChecksPerInterval: 10,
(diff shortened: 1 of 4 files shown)


pushchannelmonitor/pushchannelmonitor_test.go
```diff
@@ -315,11 +373,12 @@ func (m *mockMonitorAPI) callSubscriber(e datatransfer.Event, state datatransfer
...
-		select {
-		case <-m.restarted:
-		default:
-			close(m.restarted)
-		}
+		//select {
+		//case <-m.restarted:
+		//default:
+		//	close(m.restarted)
+		//}
```
nonsense, line 380:
```suggestion
		//select {
		//case <-m.restarted:
		//default:
		//	close(m.restarted)
		//}
```

nonsense (approved):

PR #131, opened 2021-01-14
add logging to push channel monitor

dirkmc:


nonsense:
lgtm, i'd only also add a log line at the beginning of `func (mc *monitoredChannel) checkDataRate() {	` with `mc.chid`, so that we know this is also called (also we won't have so many data channels, so probably won't be spammy.

nonsense (approved):

PR #135, opened 2021-01-18
Better channel not found error

dirkmc:


channels/channels_test.go
```diff
@@ -342,6 +343,11 @@ func TestChannels(t *testing.T) {
 	})
 }
 
+func isNotFoundError(err error) bool {
```
nonsense, line 346: I think `xerrors` contains wrapping functionality for errors, so we wouldn't have to add hacks like this and do string comparison to check if an error belongs to a certain group - in this case the `NotFound` error.

dirkmc:
Superseded by https://github.com/filecoin-project/go-data-transfer/pull/137

PR #140, opened 2021-01-26
Dont double-count queued / sent / received data when channel restarted

dirkmc:
The data transfer module subscribes to hooks called by graphsync when a block is queued or sent.
When a data transfer is restarted, Graphsync calls these hooks again with all the blocks that have already been queued and sent, with a BlockSizeOnWire() of zero.

This PR modifies event handling such that
- `DataSent` and `DataQueued` are only fired if `block.BlockSizeOnWire()` is non-zero
- `DataQueuedProgress`, `DataSentProgress` and `DataReceivedProgress` are fired only the first time a block is queued / sent / received
- The data transfer channel Queued, Sent and Received values are only increased for new blocks. eg if a block is sent twice it is not added to these values the second time.

In order to keep track of which blocks have been queued / sent / received, a new class has been added: `CIDSetManager`.
Each channel has a cid set for queued / sent / received.

### DIFF
--- a/transport/graphsync/graphsync.go
+++ b/transport/graphsync/graphsync.go
@@ -26,11 +26,6 @@ type graphsyncKey struct {
 	p         peer.ID
 }
 
-type responseProgress struct {
-	currentSent uint64
-	maximumSent uint64
-}
-
 var defaultSupportedExtensions = []graphsync.ExtensionName{extension.ExtensionDataTransfer1_1, extension.ExtensionDataTransfer1_0}
 
 // Option is an option for setting up the graphsync transport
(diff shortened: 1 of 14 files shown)


transport/graphsync/graphsync.go
```diff
@@ -400,6 +393,15 @@ func (t *Transport) gsIncomingBlockHook(p peer.ID, response graphsync.ResponseDa
 }
 
 func (t *Transport) gsBlockSentHook(p peer.ID, request graphsync.RequestData, block graphsync.BlockData) {
+	// When a data transfer is restarted, the requester sends a list of CIDs
+	// that it already has. Graphsync calls the sent hook for all blocks even
+	// if they are in the list (meaning, they aren't actually sent over the
+	// wire). So here we check if the block was actually sent
```
hannahhoward, line 399: Given your CIDsets functionality this check is probably not necessary, FWIW.

hannahhoward: It may make for greater optimization though.

hannahhoward (approved):
LGTM

PR #144, opened 2021-02-12
Fix Unmarshall of undefined ChannelID

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/143

### DIFF
--- a/types.go
+++ b/types.go
@@ -1,6 +1,7 @@
 package datatransfer
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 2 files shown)


nonsense:
@dirkmc is there a problem with making `ChannelID` a pointer? Apparently a missing `ChannelID` is a valid state, and we were just not representing it until now correctly...

dirkmc:
Yeah I guess that's true, it should really be a pointer rather than an Undef
Let me see if I can fix that with a migration..

dirkmc:
Closing in favour of https://github.com/filecoin-project/go-data-transfer/issues/143

Issue #142, opened 2021-02-11
Better error message for "incomplete response"


dirkmc:
Currently when the data transfer fails, eg because the miner crashes, the client receives the message "incomplete response".
This comes from `OnChannelCompleted()` https://github.com/filecoin-project/go-data-transfer/blob/master/impl/events.go#L319

We should provide an error message that is more human-readable, like "the sender aborted the data transfer"

Example where this has caused confusion: https://github.com/filecoin-project/lotus/issues/5409#issuecomment-766543161

PR #145, opened 2021-02-12
Better error message on complete

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/142

### DIFF
--- a/impl/events.go
+++ b/impl/events.go
@@ -316,7 +316,7 @@ func (m *manager) OnChannelCompleted(chid datatransfer.ChannelID, completeErr er
 	}
 	// send an error, but only if we haven't already errored for some reason
 	if chst.Status() != datatransfer.Failing && chst.Status() != datatransfer.Failed {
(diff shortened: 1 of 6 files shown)


impl/events.go
```diff
@@ -316,7 +316,9 @@ func (m *manager) OnChannelCompleted(chid datatransfer.ChannelID, success bool)
 	}
 	// send an error, but only if we haven't already errored for some reason
 	if chst.Status() != datatransfer.Failing && chst.Status() != datatransfer.Failed {
-		return m.channels.Error(chid, datatransfer.ErrIncomplete)
+		err := xerrors.Errorf("data transfer channel %s failed to transfer data: %w", completeErr)
```
nonsense, line 319: This seems to be missing an argument, we have %s and %w, but just completeErr as arg.

nonsense (approved):

impl/events.go
dirkmc: Thanks, good catch 👍

PR #163, opened 2021-03-18
Add ChannelStages to keep track of history of lifecycle of a DataTransfer

nonsense:


channels/channel_state.go
```diff
@@ -171,6 +173,8 @@ func (c channelState) OtherPeer() peer.ID {
 	return c.sender
 }
 
+func (c channelState) Stages() *datatransfer.ChannelStages { return c.stages }
```
dirkmc, line 176: Let's return an empty ChannelStages here if it's nil

types.go
```diff
@@ -132,4 +135,57 @@ type ChannelState interface {
...
+
+	Stages() *ChannelStages
+}
+
+type ChannelStages struct {
+	Stages []*ChannelStage
+}
+
+func (cs *ChannelStages) AddLog(stage, msg string) {
+	//log.Infof("adding log for stage <%s> msg <%s>", stage, msg)
```
dirkmc, line 147: Let's add a check for nil here and return early if cs == nil

types.go
```diff
@@ -132,4 +135,57 @@ type ChannelState interface {
...
+
+	st.Name = stage
+	st.UpdatedTime = now
+	if msg != "" && (len(st.Logs) == 0 || st.Logs[len(st.Logs)-1].Log != msg) {
+		st.Logs = append(st.Logs, &Log{msg, now})
+	}
+}
+
+func (cs *ChannelStages) GetStage(stage string) *ChannelStage {
+	for _, s := range cs.Stages {
```
dirkmc, line 166: nil check here too

channels/channel_state.go
```diff
@@ -171,6 +173,14 @@ func (c channelState) OtherPeer() peer.ID {
 	return c.sender
 }
 
+func (c channelState) Stages() *datatransfer.ChannelStages {
+	if c.stages == nil {
+		return &datatransfer.ChannelStages{}
```
raulk, line 178: This behaviour a bit footgunny. This method returns a mutable `ChannelStages` object, either from the state (which means that changes will be persisted), or an entirely new instance without setting it in the internal state (which means that mutations will be discarded).

raulk: It looks like we did it this way to save nil-checks on the user side, especially for transfers that were already in progress before this code got deployed. I'll come up with a different way.

raulk (changes requested):

channels/channel_state.go
raulk: I ended up adding a note to the godocs that mutating the return value is unsafe. There's only so much you can do with Go. Even if we returned a value instead of a pointer, the inner slices can be mutated.

raulk (approved):

PR #164, opened 2021-03-19
Demonstrate validation response issue

tchardin:
### This PR duplicates a test case from the integration tests to reproduce what I believe to be a bug:
- In this retrieval flow we don't pause the transfer when accepting the transfer as we don't need to unseal or find the content.
As a result we show that in this scenario, the response voucher does not reach the client and the first voucher they receive is the revalidation voucher request. Removing the revalidation reponse does allow the response voucher to reach the client. The expected behavior would be that a response voucher would reach the client first followed by the first revalidation voucher request.
- I was able to trace back the issue to the graphsync extension flow, when the RequestReceived hook that sends the response to the extension data hook is immediately followed by the OutgoingBlock hook that sends the first revalidation response message to the extension data hook, somehow that first message gets lost. It does go through if the OutgoingBlock hook doesn't send its message.
- This is a major blocker as it prevents usage of the data transfer protocol in the case where a retrieval acceptance does not require pausing the transfer initially before continuing.

### DIFF
--- a/impl/validation_test.go
+++ b/impl/validation_test.go
@@ -1,0 +1,208 @@
+package impl
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+
+	datatransfer "github.com/filecoin-project/go-data-transfer"
+	"github.com/filecoin-project/go-data-transfer/encoding"
+	"github.com/filecoin-project/go-data-transfer/testutil"
+	chunker "github.com/ipfs/go-ipfs-chunker"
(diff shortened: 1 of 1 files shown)


hannahhoward:
@tchardin Good catch!

 Graphsync can only send one message per extension in a response. It is up to the calling library to merge data for an extension if it needs to send multiple messages. I think perhaps the simplest move is to use a difference graphsync extension name in data transfer for the messages sent in outgoing blocks as opposed to incoming request hook. (this would be a change to the Graphsync Transport implementation in this library). @dirkmc happy to sync up about implementation cause I no longer am heavily involved with updating go-data-transfer.

PR #169, opened 2021-03-22
Use time-based counter instead of stored counter for transfer ID

dirkmc:
Currently the data transfer ID is produced by an incrementing counter stored in the datastore.

The datastore state sometimes gets out of sync in real world clients. If the state gets wiped on the client, the counter starts from zero. However miners do not accept a transfer which has an existing client peer ID and transfer ID. So all transfers from that client will be rejected until the counter reaches the same level it was at before.

To avoid this problem, this PR changes the way that the transfer ID is generated: the ID is derived from the current time.


impl/utils.go
```diff
@@ -32,11 +33,7 @@ var resumeTransportStatesResponder = statusList{
 
 // newRequest encapsulates message creation
 func (m *manager) newRequest(ctx context.Context, selector ipld.Node, isPull bool, voucher datatransfer.Voucher, baseCid cid.Cid, to peer.ID) (datatransfer.Request, error) {
-	next, err := m.storedCounter.Next()
-	if err != nil {
-		return nil, err
-	}
-	tid := datatransfer.TransferID(next)
+	tid := datatransfer.TransferID(rand.Uint64())
```
dirkmc, line 36: This is where the transfer ID is generated

nonsense (approved):
LGTM, I think this is safe.

Issue #173, opened 2021-03-24
Add config to enable / disable data-rate monitoring for push and pull channels


dirkmc:
To give clients of the library more flexibility over data rate monitoring, we should add configuration flags to allow the client to enable only push or only pull channel data rate monitoring.

PR #174, opened 2021-03-24
add config options to enable / disable push or pull monitoring individually

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/173

### DIFF
--- a/channelmonitor/channelmonitor_test.go
+++ b/channelmonitor/channelmonitor_test.go
@@ -59,6 +59,7 @@ func TestPushChannelMonitorAutoRestart(t *testing.T) {
 			mockAPI := newMockMonitorAPI(ch, tc.errOnRestart)
 
 			m := NewMonitor(mockAPI, &Config{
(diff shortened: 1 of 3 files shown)


channelmonitor/channelmonitor_test.go
```diff
@@ -550,7 +558,7 @@ func TestChannelMonitorTimeouts(t *testing.T) {
 					mch := m.AddPushChannel(ch1).(*monitoredPushChannel)
 					chCtx = mch.ctx
 				} else {
-					mch := m.AddPushChannel(ch1).(*monitoredPushChannel)
```
nonsense, line 553: How did that work before?

nonsense (approved):

channelmonitor/channelmonitor_test.go
dirkmc: It didn't 😆 
It tests the right thing now and still works so that's good news 🎉

PR #186, opened 2021-04-01
Support no-op error responses 

aarshkshah1992:
For https://github.com/filecoin-project/go-fil-markets/issues/528.

@dirkmc 

As discussed, we need to add support for `go-fil-markets` to return an `ErrNoOp` in addition to `ErrPause` & `ErrResume`so data transfer can maintain the status quo instead of resuming/pausing the data transfer to avoid races with the `go-fil-markets` resuming data transfer in the state machine thread.


aarshkshah1992:
@dirkmc 

```
=== RUN   TestChannelMonitorTimeouts/pull:_complete_too_late

```

fails CI.

Is there known flakiness in the Channel Monitor tests ?

dirkmc:
> Is there known flakiness in the Channel Monitor tests?

Yes, that's a known issue, we can safely ignore 👍

errors.go
```diff
@@ -25,6 +25,10 @@ const ErrPause = errorType("pause channel")
 // use to resume the channel
 const ErrResume = errorType("resume channel")
 
+// ErrNoOp is a special error that the RequestReceived / ResponseReceived hooks can
+// use to neither resume nor pause the channel but maintain the status quo.
+const ErrNoOp = errorType("no-op")
+
```
dirkmc, line 31: Instead of introducing ErrNoOp, would it make sense to change the code to
- explicitly return ErrResume when the revalidator wants to resume the transfer
- return nil when the revalidator wants a NoOp

aarshkshah1992: @dirkmc 

I thought of that but doing that breaks a whole bunch of tests in `go-data-transfer`. Happy to go down this path and fix the tests if you think that's more valuable.

dirkmc (approved):
LGTM 👍

PR #182, opened 2021-03-29
Enable channel data-rate monitoring separately from creating channel

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/172

TODO:
- [x] Implementation
- [x] Unit tests
- [x] Integration test

PR #181, opened 2021-03-29
Respect pausing in pull channel monitor

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/177

Issue #177, opened 2021-03-25
Respect pausing in data transfer channel data-rate monitors


dirkmc:
During retrieval, the provider sends data and the client sends payment vouchers. The provider sends a "pause" when waiting for payment.

If the client needs to add funds to the payment channel it can take a few epochs for the funds to be confirmed on chain.

When the channel is paused, the channel monitor should stop checking the data rate.

dirkmc:
Superseded by https://github.com/filecoin-project/go-fil-markets/pull/540

Issue #172, opened 2021-03-23
Add mechanism to start push / pull channel data rate monitoring 


dirkmc:
Currently the push or pull channel data rate monitoring starts automatically when the channel is opened.

However in go-fil-markets for retrieval [we open a pull channel](https://github.com/filecoin-project/go-fil-markets/blob/master/retrievalmarket/impl/clientstates/client_states.go#L28), then create a payment channel. It can take a long time to open a payment channel which means the data-rate will be zero, and the pull channel monitor will try to restart the connection.

Instead we should add a mechanism to manually start data-rate monitoring at some point after the channel has been opened.

dirkmc:
Fixed by https://github.com/filecoin-project/go-fil-markets/pull/540

PR #190, opened 2021-04-06
Channel monitor watches for errors instead of measuring data rate

dirkmc:
Depends on https://github.com/ipfs/go-graphsync/pull/164
Supercedes https://github.com/filecoin-project/go-data-transfer/issues/177 https://github.com/filecoin-project/go-data-transfer/issues/172 https://github.com/filecoin-project/go-data-transfer/pull/182 https://github.com/filecoin-project/go-data-transfer/pull/181

Currently the channel monitor watches the data-rate and if it falls too low, the monitor sends a Restart message to the other party in the data transfer.

This is complex to do for retrieval deals, because the monitor needs to be aware of pauses and resumes.

Instead this PR simplifies the monitor's behaviour so that now:
- The monitor doesn't watch the data-rate
- The monitor just watches for errors fired from the transport layer
- When there is an error, the monitor tries to reconnect to the peer
- Once the connection is established, the monitor sends a restart message
- If there is no response to the restart message, the data transfer fails

### DIFF
--- a/channelmonitor/channelmonitor.go
+++ b/channelmonitor/channelmonitor.go
@@ -489,7 +489,7 @@ func (mc *monitoredChannel) closeChannelAndShutdown(cherr error) {
 // Wait for the peer to send an acknowledgement to the restart request
 func (mc *monitoredChannel) waitForRestartResponse() chan error {
 	restartFired := make(chan struct{})
-	restarted := make(chan error, 3)
+	restarted := make(chan error, 1)
 	timer := time.NewTimer(mc.cfg.RestartAckTimeout)
 
 	unsub := mc.mgr.SubscribeToEvents(func(event datatransfer.Event, channelState datatransfer.ChannelState) {
--- a/channelmonitor/channelmonitor.go
+++ b/channelmonitor/channelmonitor.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"golang.org/x/xerrors"
 
 	datatransfer "github.com/filecoin-project/go-data-transfer"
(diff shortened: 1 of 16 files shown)


network/libp2p_impl.go
```diff
@@ -187,6 +187,18 @@ func (dtnet *libp2pDataTransferNetwork) ConnectTo(ctx context.Context, p peer.ID
...
+// OpenStreamTo establishes a connection to the given peer, retrying if
+// necessary, and opens a stream on the data-transfer protocol to verify
+// the peer will accept messages on the protocol
+func (dtnet *libp2pDataTransferNetwork) OpenStreamTo(ctx context.Context, p peer.ID) error {
+	s, err := dtnet.openStream(ctx, p, dtnet.dtProtocols...)
+	if err != nil {
+		return err
+	}
+
+	return s.Close()
```
nonsense, line 199: I am not sure why `s.Close()` is returned immediately and not in a `defer func`.. we don't need the `stream` or?

nonsense (approved):

channelmonitor/channelmonitor.go
```diff
@@ -321,52 +328,94 @@ func (mc *monitoredChannel) isRestarting() bool {
...
+	log.Infof("%s: restarting (%d consecutive restarts)", mc.chid, restartCount)
 	err := mc.sendRestartMessage(restartCount)
 	if err != nil {
-		// If the restart message could not be sent, close the channel and
-		// shutdown the monitor
-		mc.closeChannelAndShutdown(err)
-		return
+		log.Warnf("%s: restart failed, trying again: %s", mc.chid, err)
+		// If the restart message could not be sent, or there was a timeout
+		// waiting for the restart to be acknowledged, try again
```
nonsense, line 413: Shouldn't we have some kind of a sleep here? Seems like we can end up firing a lot of consecutive messages.

channelmonitor/channelmonitor.go
```diff
@@ -440,7 +489,7 @@ func (mc *monitoredChannel) closeChannelAndShutdown(cherr error) {
 // Wait for the peer to send an acknowledgement to the restart request
 func (mc *monitoredChannel) waitForRestartResponse() chan error {
 	restartFired := make(chan struct{})
-	restarted := make(chan error)
+	restarted := make(chan error, 3)
```
nonsense, line 492: Buffered chan with size 3? Why is that? I guess it makes sense that we don't want to block when we push to `restarted`, but why 3? Isn't it possible to block when we push to the channel 3 times?

channelmonitor/channelmonitor.go
dirkmc: `sendRestartMessage` tries to connect to the peer, and then waits for a restart ack.
When connecting to the peer, it has retries with backoff. If it returns an error it will be after having tried several times.
If it returns an error waiting for a restart ack it will be after a timeout.

So in both of these cases there is an implied delay before it returns an error.

channelmonitor/channelmonitor.go
dirkmc: We should only get a single restart event, but potentially we could also get a context timeout and a timer expiry.
But actually as soon as we get any of those events we exit the select to it doesn't make sense for it to be 3, I will fix

PR #191, opened 2021-04-07
Race between "Completed" event and reading the blocks from the store

tchardin:
#### This PR is for demonstration purposes only:
I add a unit test in which a single client sends a message to multiple providers which upon reception all pull the same data from the client. Transfers are successful however when the data receivers try to read the file after they receive a "Completed" status in the event subscriber, they get a `failed to fetch all nodes` error. It seems to show that receiving a "Completed" event doesn't guarantee the blocks are ready to read/  It works fine when adding a 1 second delay in between or running with `-race` flag.
To reproduce simply checkout this branch, cd in /impl and run `go test -v -run=MultiRequestStreams .`


dirkmc:
Superseded by https://github.com/filecoin-project/go-data-transfer/pull/205

PR #195, opened 2021-04-14
Prevent concurrent restarts for same channel

dirkmc:
This PR makes sure that when the data transfer is restarted for a pull request (retrieval), it cannot be interrupted by a second concurrent restart request for the same data transfer channel.

### DIFF
--- a/transport/graphsync/graphsync.go
+++ b/transport/graphsync/graphsync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-graphsync"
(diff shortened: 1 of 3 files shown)


transport/graphsync/graphsync.go
```diff
@@ -21,6 +22,13 @@ var log = logging.Logger("dt_graphsync")
 
 var errContextCancelled = errors.New("context cancelled")
 
+// When restarting a data transfer, we cancel the existing graphsync request
+// before opening a new one.
+// These constants define the minimum and maximum time to wait for the request
+// to be cancelled.
+const minGSCancelWait = 100 * time.Millisecond
+const maxGSCancelWait = time.Second
```
nonsense, line 30: Overall LGTM, I am just not sure how we decide that these values are reasonable...

nonsense (approved):

transport/graphsync/graphsync.go
dirkmc: It's true, ideally there would be a way to ask graphsync to cancel a request and wait until all events have been emitted before returning

PR #199, opened 2021-04-28
Fix pause on restart

dirkmc:
When there is a restart, and the data transfer hasn't yet started (eg because the data is still being unsealed) the new graphsync request should start off in the paused state

### DIFF
--- a/channelmonitor/channelmonitor.go
+++ b/channelmonitor/channelmonitor.go
@@ -52,6 +52,8 @@ type Config struct {
 	// Max time to wait for the responder to send a Complete message once all
 	// data has been sent
 	CompleteTimeout time.Duration
+	// Called when a restart completes successfully
+	OnRestartComplete func(id datatransfer.ChannelID)
 }
 
 func NewMonitor(mgr monitorAPI, cfg *Config) *Monitor {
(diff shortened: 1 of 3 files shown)


channelmonitor/channelmonitor.go
```diff
@@ -382,6 +384,9 @@ func (mc *monitoredChannel) restartChannel() {
 
 		if !restartAgain {
 			// No restart queued, we're done
+			if mc.cfg.OnRestartComplete != nil {
+				mc.cfg.OnRestartComplete(mc.chid)
```
aarshkshah1992, line 388: I don't see this getting configured/injected anywhere. What's the purpose of this ?

transport/graphsync/graphsync.go
```diff
@@ -375,10 +384,11 @@ func (t *Transport) CloseChannel(ctx context.Context, chid datatransfer.ChannelI
 		return nil
 	}
 	t.dataLock.Lock()
-	if _, ok := t.requestorCancelledMap[chid]; ok {
+	_, ok := t.requestorCancelledMap[chid]
+	t.dataLock.Unlock()
+	if ok {
```
aarshkshah1992, line 389: Woah. Great catch man ! This could have been causing deadlock issues ...

aarshkshah1992 (approved):
@dirkmc Some quick questions. Otherwise LGTM.

channelmonitor/channelmonitor.go
dirkmc: This is just used by the tests

transport/graphsync/graphsync.go
dirkmc: Yeah I think it actually *has* caused some deadlock issues, I suspect once we release we'll see a bunch of weird problems go away.

PR #204, opened 2021-05-11
feat: use different extension names to fit multiple hooks data in same graphsync message

tchardin:
As suggested by @hannahhoward, this PR uses different extension names for the income request and outgoing blocks so they can be included in the same extension message. Not sure how to handle the `SupportedExtensions` option in the transport constructor, this is rather hacky but it unblocks use cases in which request validation does not pause the manager for unsealing.

### DIFF
--- a/impl/integration_test.go
+++ b/impl/integration_test.go
@@ -3,6 +3,8 @@ package impl_test
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"io"
 	"math/rand"
 	"os"
 	"testing"
@@ -18,9 +20,13 @@ import (
 	gsnet "github.com/ipfs/go-graphsync/network"
 	"github.com/ipfs/go-graphsync/storeutil"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
+	chunker "github.com/ipfs/go-ipfs-chunker"
 	offline "github.com/ipfs/go-ipfs-exchange-offline"
+	files "github.com/ipfs/go-ipfs-files"
 	ipldformat "github.com/ipfs/go-ipld-format"
 	"github.com/ipfs/go-merkledag"
+	"github.com/ipfs/go-unixfs/importer/balanced"
+	ihelper "github.com/ipfs/go-unixfs/importer/helpers"
 	"github.com/ipld/go-ipld-prime"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/libp2p/go-libp2p-core/host"
@@ -1069,6 +1075,7 @@ func TestSimulatedRetrievalFlow(t *testing.T) {
 				if channelState.Status() == datatransfer.Completed {
 					clientFinished <- struct{}{}
 				}
+				fmt.Println("=> CLIENT", datatransfer.Events[event.Code], datatransfer.Statuses[channelState.Status()])
 			}
 			dt2.SubscribeToEvents(clientSubscriber)
 			providerFinished := make(chan struct{}, 1)
@@ -1090,6 +1097,7 @@ func TestSimulatedRetrievalFlow(t *testing.T) {
 				if channelState.Status() == datatransfer.Completed {
 					providerFinished <- struct{}{}
 				}
+				fmt.Println("=> PROVIDER", datatransfer.Events[event.Code], datatransfer.Statuses[channelState.Status()])
 			}
 			dt1.SubscribeToEvents(providerSubscriber)
 			voucher := testutil.FakeDTType{Data: "applesauce"}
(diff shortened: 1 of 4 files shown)


impl/integration_test.go
```diff
@@ -1713,6 +1718,163 @@ func TestRespondingToPullGraphsyncRequests(t *testing.T) {
...
+		if event.Code == datatransfer.DataReceived &&
+			clientPausePoint < len(pausePoints) &&
+			channelState.Received() > pausePoints[clientPausePoint] {
+			_ = dt2.SendVoucher(ctx, chid, testutil.NewFakeDTType())
+			clientPausePoint++
+		}
+
+		if channelState.Status() == datatransfer.Completed {
+			clientFinished <- struct{}{}
+		}
```
dirkmc, line 1786: Could we add some comments here to explain what is being verified

impl/integration_test.go
```diff
@@ -1713,6 +1718,163 @@ func TestRespondingToPullGraphsyncRequests(t *testing.T) {
...
+	clientGotResponse := make(chan struct{}, 1)
+
+	// The response voucher is the voucher returned by the request validator
+	respVoucher := testutil.NewFakeDTType()
+	encodedRVR, err := encoding.Encode(respVoucher)
+	require.NoError(t, err)
+
+	finalVoucherResult := testutil.NewFakeDTType()
+	encodedFVR, err := encoding.Encode(finalVoucherResult)
+	require.NoError(t, err)
```
dirkmc, line 1760: Could we add some comments here to explain what each of these vouchers are any why we need them

impl/integration_test.go
```diff
@@ -1713,6 +1718,163 @@ func TestRespondingToPullGraphsyncRequests(t *testing.T) {
...
+	gsData.VerifyFileTransferred(t, root, true)
+}
+
+func LoadRandomData(ctx context.Context, t *testing.T, dagService ipldformat.DAGService) (ipld.Link, []byte) {
+	tf, err := os.CreateTemp("", "data")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		os.Remove(tf.Name())
+	})
+
```
dirkmc, line 1842: Is it possible to create random data in memory, instead of creating a temp file?

transport/graphsync/extension/gsextension.go
```diff
@@ -71,7 +77,15 @@ func GetTransferData(extendedData GsExtended) (datatransfer.Message, error) {
...
-			return nil, nil
+			extName = ExtensionIncomingRequest1_1
+			data, ok = extendedData.Extension(extName)
+			if !ok {
+				extName = ExtensionOutgoingBlock1_1
+				data, ok = extendedData.Extension(extName)
+				if !ok {
+					return nil, nil
+				}
+			}
```
dirkmc, line 88: nit: can we use a loop here?

transport/graphsync/extension/gsextension.go
```diff
@@ -14,6 +14,10 @@ import (
 )
 
 const (
+	// ExtensionIncomingRequest1_1 is the identifier for data sent by the IncomingReequest hook
```
dirkmc, line 17:
```suggestion
	// ExtensionIncomingRequest1_1 is the identifier for data sent by the IncomingRequest hook
```
```suggestion
	// ExtensionIncomingRequest1_1 is the identifier for data sent by the IncomingReequest hook
```

dirkmc (commented):
Thanks for making this PR @tchardin

dirkmc:
@hannahhoward I'm not so familiar with how the extensions work, could you please take a look as well.

Is this going to require a protocol version number change?

transport/graphsync/extension/gsextension.go
```diff
@@ -65,14 +71,22 @@ type GsExtended interface {
...
+	i := 0
+	extName := extNames[i]
 	data, ok := extendedData.Extension(extName)
-	if !ok {
-		extName = ExtensionDataTransfer1_0
-		data, ok = extendedData.Extension(extName)
-		if !ok {
+	for !ok {
+		i++
+		if i == len(extNames) {
```
dirkmc, line 85: One more nit - can we restructure this for loop a little, eg:
```
var ok bool
var []byte data
for i := 0; i < len(extNames) && !ok; i++ {
  data, ok = extendedData.Extension(extNames[i])
}
if !ok {
  return nil, nil
}
```

tchardin: Good call thanks!

dirkmc: Looks good 👍

dirkmc:
@tchardin I did some refactoring in the graphsync transport to make it less racy - could you please rebase on top of master 🙏

transport/graphsync/graphsync.go
```diff
@@ -635,12 +660,23 @@ func (t *Transport) gsIncomingResponseHook(p peer.ID, response graphsync.Respons
 	if err != nil {
 		hookActions.TerminateWithError(err)
 	}
+
+	// In a case where the transfer sends blocks immediately this extension may contain both a
+	// response message and a revalidation request so we trigger OnResponseReceived again for this
+	// specific extension name
+	msg, err := extension.GetTransferData(response, []graphsync.ExtensionName{extension.ExtensionOutgoingBlock1_1})
```
hannahhoward, line 667: Nit: you should use processExtensions here with the list above. We can just ignore ResponseMessage I think for now?

hannahhoward (approved):
Couple nits but generally LGTM. Really awesome you did this on your own. I just got back from vacation and was planning to do this one, only to find it done!

hannahhoward (approved):

PR #205, opened 2021-05-12
Completed race + fix for test

dirkmc:
This PR adds a commit to https://github.com/filecoin-project/go-data-transfer/pull/191.

The problem was that the test was waiting for all data to be _sent_. It would listen to events and wait for the channel to reach the Completed status on the _sending_ side.

The test then tried to verify that all data had been received on the _receiving_ side. But there is a small delay between sending data and receiving data.

So the fix in [commit 04d38cd](https://github.com/filecoin-project/go-data-transfer/commit/04d38cd8b706bd3e72c32902a3bc01347e877b49) is to modify the test to watch for the Complete status on the _receiving_ side, then to verify that all data has been received.


dirkmc:
@tchardin I'm going to close this PR, please try it out and reopen if you're still seeing an issue.

Issue #166, opened 2021-03-19
investigate low performance and CPU hotspot transferring 4GiB over loopback


raulk:
This is an initial observation from running the Whitenoise tests with a perfect network and no interruption rate, over the loopback interface. In other words, these are ideal conditions for a data transfer.

The total payload size that was transfered was 4GiB.

The transfer stack was go-data-transfer with graphsync, conducting a push from the sender to the receiver.

The hardware used was a MBP 16", 2,3 GHz 8-Core Intel Core i9, 16 GB 2667 MHz DDR4.

* The import of the random 4GiB payload into the badger store took 22109ms.
* The transfer as perceived from the client took 117000ms (almost 2 minutes).
* The average data transfer rate is 35MB/s -- this is very poor for loopback.

Attached is the run result. **IMPORTANT: this is really a gzip, but GitHub only allows uploading zips: change the extension to gzip!**

[c1ah45gi7qkqo2qcbicg.zip](https://github.com/filecoin-project/go-data-transfer/files/6174136/c1ah45gi7qkqo2qcbicg.zip)

Of special interest are the `events` files, which capture the subscription stream from go-data-transfer from both the sender and receiver. Very useful for debugging.

There are heap profiles and CPU profiles included too (using Testground's new [profiling feature](https://docs.testground.ai/v/master/analyzing-the-results/capturing-profiles)) for further digging.

So far, I found what appears to be a large hotspot on the receiving side. Addressing this hotspot _could_ allow the receiver to ACK data sooner, and thus speed up the data flow.

The hotspot seems to be in the "CID lists" feature, which I'm not familiar with.

![profile001](https://user-images.githubusercontent.com/1101242/111844861-b9c81b00-88fb-11eb-934f-365a9a56c39a.png)

NOTE: If you run the test, make sure you use the `feat/tmp-dir` branch of Testground, as Whitenoise depends on the [new temporary directory feature](https://github.com/testground/testground/pull/1212). This is also noted in the Whitenoise README.

dirkmc:
When go-data-transfer receives a block, it adds the CID to a list.
If the data transfer is restarted, eg because the connection went down, the receiving side sends a list of all CIDs that it has received to the sender, so that the sender can skip over these CID and only resend missing CIDs to the receiver.

[...]

PR #212, opened 2021-06-01
Update go-graphsync v0.8.0

hannahhoward:
# Goals

Good news: The latest version of go-graphync and go-ipld-prime contains an important optimization to remove a bunch of unneccesary hashing when reading from local blockstores, which should cut down significantly on CPU usage
Bad news: The latest versions of these libraries contain a some important, but very breaking changes to syntax.
Good news: I'm updating them all the way to Lotus! And also, in most cases, from the point of view of the consuming libraries, they're simplifications.

# Implementation

- Update to go-graphsync v0.8.0 with go-ipld-prime linksystem branch & trusted store.
- Fix syntax changes which are mostly to:
   - dagcbor Encode / Decode function name
   - use of linksystem over loader + storer.
   
# Background

- More about the LinkSystem: https://github.com/ipld/go-ipld-prime/pull/143
- Trusted store branch: https://github.com/ipld/go-ipld-prime/pull/149
- PR to merge LinkSystem into go-graphsync: https://github.com/ipfs/go-graphsync/pull/161


tchardin:
@dirkmc what is the timing for this? Unfortunately we cannot use the latest go-data-transfer release because it relies on changes made to an older version of graphsync that is incompatible with ipld-prime v0.7.0 used in our project. Cannot run it from the commit either since there are breaking changes with go-fil-markets. So sadly we can't benefit from the changes you made until this gets in.

dirkmc:
@tchardin we've been doing some performance testing of graphsync v0.8.0. You can follow the progress at https://github.com/filecoin-project/go-data-transfer/issues/166

It seems like there may need to be further optimizations made before a newer version of graphsync can be integrated into go-data-transfer.

I'd suggest you align your code with the [latest release of go-data-transfer v1.7.0](https://github.com/filecoin-project/go-data-transfer/releases/tag/v1.7.0) which uses [graphsync v0.6.4](https://github.com/filecoin-project/go-data-transfer/blob/v1.7.0/go.mod#L16)

tchardin:
Gotcha, thanks @dirkmc. So you confirm it would be faster to downgrade our project to ipld-prime v0.5.1 instead of waiting for this to get in?

dirkmc:
I think that's probably your best bet 👍

PR #216, opened 2021-06-03
Store the count of unique Receieved cids

aarshkshah1992:
This PR introduces the _minimal_ set of changes we need to fix the [logging hotspot ](https://github.com/filecoin-project/lotus/blob/master/markets/loggers/loggers.go#L43)in Lotus that happens because of reading the entire list of received cids from disk.

Lotus PR at:
https://github.com/filecoin-project/lotus/pull/6385

@dirkmc I've made this PR against master but we need to merge it into v1.6.0 as well and release to avoid the latest changes in master from making their way to these fixes, right ?


dirkmc:
Closing in favour of https://github.com/filecoin-project/go-data-transfer/pull/217

PR #227, opened 2021-07-29
Cleanup channel in a go-routine

aarshkshah1992:
`go-statemachine` dosen't process additional events till this state handler function executes. We should do the cleanup in a go-routine.


hannahhoward (commented):
do we have any evidence this is ever taking a long time? or is this just a precaution?

Asking cause I think we need to be careful about logic changes this close to M1 unless we believe for sure that it is related to the most recent issues.

PR #229, opened 2021-08-05
Simplify graphsync cancel

dirkmc:
Depends on https://github.com/ipfs/go-graphsync/pull/188

### DIFF
--- a/channels/channels_fsm.go
+++ b/channels/channels_fsm.go
@@ -95,9 +95,9 @@ var ChannelEvents = fsm.Events{
 		chst.AddLog("data transfer receive error: %s", chst.Message)
 		return nil
 	}),
-	fsm.Event(datatransfer.RequestTimedOut).FromAny().ToNoChange().Action(func(chst *internal.ChannelState, err error) error {
+	fsm.Event(datatransfer.RequestCancelled).FromAny().ToNoChange().Action(func(chst *internal.ChannelState, err error) error {
 		chst.Message = err.Error()
-		chst.AddLog("data transfer request timed out: %s", chst.Message)
+		chst.AddLog("data transfer request cancelled: %s", chst.Message)
 		return nil
 	}),
 	fsm.Event(datatransfer.Error).FromAny().To(datatransfer.Failing).Action(func(chst *internal.ChannelState, err error) error {
--- a/transport.go
+++ b/transport.go
@@ -56,9 +56,10 @@ type EventsHandler interface {
 	// Error returns are logged but otherwise have no effect
 	OnChannelCompleted(chid ChannelID, err error) error
 
-	// OnRequestTimedOut is called when a request we opened (with the given channel Id) to receive data times out.
+	// OnRequestCancelled is called when a request we opened (with the given channel Id) to
+	// receive data is cancelled by us.
 	// Error returns are logged but otherwise have no effect
-	OnRequestTimedOut(chid ChannelID, err error) error
+	OnRequestCancelled(chid ChannelID, err error) error
 
 	// OnRequestDisconnected is called when a network error occurs trying to send a request
 	OnRequestDisconnected(chid ChannelID, err error) error
--- a/transport/graphsync/graphsync.go
+++ b/transport/graphsync/graphsync.go
@@ -897,7 +897,8 @@ func (c *dtChannel) gsReqOpened(gsKey graphsyncKey, hookActions graphsync.Outgoi
 }
 
 // gsDataRequestRcvd is called when the transport receives an incoming request
-// for data
+// for data.
+// Note: Must be called under the lock.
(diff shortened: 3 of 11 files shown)


channels/channels_fsm.go
```diff
@@ -95,9 +95,9 @@ var ChannelEvents = fsm.Events{
 		chst.AddLog("data transfer receive error: %s", chst.Message)
 		return nil
 	}),
-	fsm.Event(datatransfer.RequestTimedOut).FromAny().ToNoChange().Action(func(chst *internal.ChannelState, err error) error {
+	fsm.Event(datatransfer.RequestCancelled).FromAny().ToNoChange().Action(func(chst *internal.ChannelState, err error) error {
```
raulk, line 98: Since we are changing this state transition and RequestTimedOut is abandoned now, should we migrate all RequestTimedOut to RequestCancelled, so that they're not left in limbo? Or are these terminal states?

Alternatively, we could rename RequestTimedOut to RequestCancelled?

transport.go
```diff
@@ -56,9 +56,10 @@ type EventsHandler interface {
 	// Error returns are logged but otherwise have no effect
 	OnChannelCompleted(chid ChannelID, err error) error
 
-	// OnRequestTimedOut is called when a request we opened (with the given channel Id) to receive data times out.
+	// OnRequestCancelled is called when a request we opened (with the given channel Id) to
+	// receive data is cancelled by us.
 	// Error returns are logged but otherwise have no effect
-	OnRequestTimedOut(chid ChannelID, err error) error
+	OnRequestCancelled(chid ChannelID, err error) error
```
raulk, line 62: Do we need to percolate this change up to go-fil-markets, or lotus?

transport/graphsync/graphsync.go
```diff
@@ -177,36 +170,28 @@ func (t *Transport) consumeResponses(responseChan <-chan graphsync.ResponseProgr
 }
 
 // Read from the graphsync response and error channels until they are closed
-// or there is an error, then call the channel completed callback
+// or there is an error
 func (t *Transport) executeGsRequest(req *gsReq) {
-	defer req.onComplete()
```
raulk, line 182: I'm not an expert in go-data-transfer, but what replaces the onComplete call here?

transport/graphsync/graphsync.go
```diff
@@ -1126,14 +1063,34 @@ func (c *dtChannel) cleanup() {
...
 	c.lk.Lock()
 	defer c.lk.Unlock()
 
 	// Cancel the graphsync request
-	if c.cancelReq != nil {
-		c.cancelReq.cancel()
+	return c.cancelRequest(ctx)
+}
+
+// Cancel the graphsync request
```
raulk, line 1074: Can we specify in the godoc that this MUST be called under the lock?

transport/graphsync/graphsync.go
```diff
@@ -981,7 +920,7 @@ func (c *dtChannel) gsDataRequestRcvd(gsKey graphsyncKey, hookActions graphsync.
 
 	// Save a mapping from the graphsync key to the channel ID so that
 	// subsequent graphsync callbacks are associated with this channel
-	c.gsKey = gsKey
+	c.gsKey = &gsKey
```
raulk, line 923: We don't seem to have acquired the lock before writing 🤔

channels/channels_fsm.go
dirkmc: RequestTimedOut & RequestCancelled are events, not states, so I don't think they will get recorded anywhere. I don't think we need to migrate anything.

transport.go
dirkmc: This interface is only used internally so I don't think we need to make changes anywhere else

transport/graphsync/graphsync.go
dirkmc: Now that `graphsync.CancelRequest()` waits until the event queue is drained, we don't need to detect complete manually so we no longer need this callback

transport/graphsync/graphsync.go
dirkmc: The lock is acquired on line 515. The code in this file is super complicated to follow.

transport/graphsync/graphsync.go
dirkmc: Actually it is safer to wait for both the events to be drained, and the channel returned from `graphsync.Request()` to be drained. So in the end I think it does make sense to keep the onComplete mechanism 👍 
I've restored it now.

transport.go
raulk: Ack.

transport/graphsync/graphsync.go
raulk: Ack.

raulk (approved):

PR #232, opened 2021-08-17
Log closing of completion channel

aarshkshah1992:
Log closing of the completion _channel_ of the data transfer channel when the corresponding Graphsync request completes.


raulk (approved):
Looks inoffensive and not chatty.

PR #185, opened 2021-04-01
Dont double count data sent

dirkmc:
Fixes https://github.com/filecoin-project/go-data-transfer/issues/184

Tasks:
- [x] Implement fix
- [x] Unit tests
- [x] Integration tests

PR #233, opened 2021-08-18
Handle data-sent and data-queued events in the TransferFinished state

### LINKED ISSUE #234: DataSent and DataQueued events should be handled gracefully in the TransferFinished state
- After a data-transfer restart, Graphsync can still call the `OnDataSent` hook for blocks that have already been sent which in turn will fire the `DataSent` data-transfer event. The same applies to the `DataQueued` event.
- To ensure we don't double count data sent, we introduced a `DataSentProgress` and a `DataQueuedProgress` events a while ago in https://github.com/filecoin-project/go-data-transfer/pull/185.  These events are only emitted the first time we send a block to a remote peer.
- Thus, if a data-transfer restarts after entering the `TransferFinished` state, we can still receive `DataSent` and `DataQueued` events for  blocks that have already been sent. We need to handle these events gracefully in the `TransferFinished` state.

aarshkshah1992:
Closes #234 .

- After a data-transfer restart, Graphsync can still call the `OnDataSent` hook for blocks that have already been sent which in turn will fire the `DataSent` data-transfer event. The same applies to the `DataQueued` event.
- To ensure we don't double count data sent, we introduced a `DataSentProgress` and a `DataQueuedProgress` events a while ago in https://github.com/filecoin-project/go-data-transfer/pull/185.  These events are only emitted the first time we send a block to a remote peer.
- Thus, if a data-transfer restarts after entering the `TransferFinished` state, we can still receive `DataSent` and `DataQueued` events for  blocks that have already been sent. We need to handle these events gracefully in the `TransferFinished` state. We do see a LOT of restarts in Estaury's logs for data-transfer across the board. See attached.
- Note however, that we should NOT see the `DataSentProgress` and a `DataQueuedProgress`  events in the `TransferFinished` state and this is confirmed by Estuary's logs.

**_Note:_** I don't think this error actually causes storage deals to fail but still nice to fix to clean up the logs.

[shuttle-logs-bad-tx3 (1).gz](https://github.com/filecoin-project/go-data-transfer/files/7005749/shuttle-logs-bad-tx3.1.gz)

### DIFF
--- a/channels/channels_fsm.go
+++ b/channels/channels_fsm.go
@@ -60,17 +60,26 @@ var ChannelEvents = fsm.Events{
 			chst.AddLog("received data")
 			return nil
 		}),
-	fsm.Event(datatransfer.DataSent).FromMany(transferringStates...).ToNoChange().Action(func(chst *internal.ChannelState) error {
+
+	fsm.Event(datatransfer.DataSent).
+		FromMany(transferringStates...).ToNoChange().
+		From(datatransfer.TransferFinished).ToNoChange().
+		Action(func(chst *internal.ChannelState) error {
 		chst.AddLog("")
 		return nil
 	}),
+
 	fsm.Event(datatransfer.DataSentProgress).FromMany(transferringStates...).ToNoChange().
 		Action(func(chst *internal.ChannelState, delta uint64) error {
 			chst.Sent += delta
 			chst.AddLog("sending data")
 			return nil
 		}),
-	fsm.Event(datatransfer.DataQueued).FromMany(transferringStates...).ToNoChange().Action(func(chst *internal.ChannelState) error {
+
+	fsm.Event(datatransfer.DataQueued).
+		FromMany(transferringStates...).ToNoChange().
+		From(datatransfer.TransferFinished).ToNoChange().
+		Action(func(chst *internal.ChannelState) error {
 		chst.AddLog("")
 		return nil
 	}),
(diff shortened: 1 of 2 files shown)


channels/channels_test.go
```diff
@@ -138,6 +138,39 @@ func TestChannels(t *testing.T) {
...
+		state := checkEvent(ctx, t, received, datatransfer.FinishTransfer)
+		require.Equal(t, datatransfer.TransferFinished, state.Status())
+
+		// send a data-sent event and ensure it's a no-op
+		_, err = channelList.DataSent(chid, cids[1], 1)
+		require.NoError(t, err)
+		state = checkEvent(ctx, t, received, datatransfer.DataSent)
+		require.Equal(t, datatransfer.TransferFinished, state.Status())
+
+
```
raulk, line 166: needs `gofmt`.

channels/channels_fsm.go
```diff
@@ -60,17 +60,26 @@ var ChannelEvents = fsm.Events{
...
 		}),
-	fsm.Event(datatransfer.DataSent).FromMany(transferringStates...).ToNoChange().Action(func(chst *internal.ChannelState) error {
+
+	fsm.Event(datatransfer.DataSent).
+		FromMany(transferringStates...).ToNoChange().
+		From(datatransfer.TransferFinished).ToNoChange().
+		Action(func(chst *internal.ChannelState) error {
 		chst.AddLog("")
 		return nil
 	}),
```
raulk, line 70: Bad indenting here and below.

raulk (approved):
CI failing due to `gofmt` issues.

channels/channels_test.go
aarshkshah1992: Apologies, need to setup gofmt on my new Goland installation.

PR #236, opened 2021-08-19
Log completion message flow

aarshkshah1992:
@raulk This is to debug the "Timed out waiting for complete message" issue.

### DIFF
--- a/impl/events.go
+++ b/impl/events.go
@@ -165,6 +165,11 @@ func (m *manager) OnTransferQueued(chid datatransfer.ChannelID) {
 }
 
 func (m *manager) OnResponseReceived(chid datatransfer.ChannelID, response datatransfer.Response) error {
+	if response.IsComplete() {
+		log.Infow("received complete response", "chid", chid, "isAccepted", response.Accepted())
+	}
+
+
 	if response.IsCancel() {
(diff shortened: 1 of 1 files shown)


impl/events.go
```diff
@@ -244,17 +250,20 @@ func (m *manager) OnChannelCompleted(chid datatransfer.ChannelID, completeErr er
 	if completeErr == nil {
 		// If the channel was initiated by the other peer
 		if chid.Initiator != m.peerID {
+			log.Infow("received OnChannelCompleted, will send completion message to initiator", "chid", chid)
 			msg, err := m.completeMessage(chid)
 			if err != nil {
-				return nil
+				return err
```
nonsense, line 256: Is this changed on purpose? I see everything else is log lines, apart from this.

nonsense (approved):
LGTM, just one comment :)

impl/events.go
raulk: This needs an explanation, as it introduces a change in behaviour.

raulk (changes requested):

impl/events.go
aarshkshah1992: Hey,

The error returned from here is only logged and we take no action based on it. If there's an error while creating the complete message, we do want to know about it.

raulk (approved):

PR #240, opened 2021-08-23
test: reproduce OnComplete race condition

tchardin:
This is reproducing an issue I've been experiencing with our Filecoin retrievals. When operating multiple transfers sequentially with single block DAGs, the last voucher result (sent via data-transfer network protocol) arrives before the acceptance voucher result (attached to the graphsync extension.


tchardin:
Was able to hack around it so will close this as I don't know if it's worth spending time on at this point.

PR #241, opened 2021-08-24
reduce channel monitor log verbosity

dirkmc:


channelmonitor/channelmonitor.go
```diff
@@ -433,17 +458,17 @@ func (mc *monitoredChannel) sendRestartMessage(restartCount int) error {
...
 	if err != nil {
 		return xerrors.Errorf("%s: failed to reconnect to peer %s after %s: %w",
 			mc.chid, p, time.Since(start), err)
 	}
-	log.Infof("%s: re-established connection to %s in %s", mc.chid, p, time.Since(start))
+	log.Debugf("%s: re-established connection to %s in %s", mc.chid, p, time.Since(start))
 
 	// Send a restart message for the channel
-	log.Infof("%s: sending restart message to %s (%d consecutive restarts)", mc.chid, p, restartCount)
+	log.Debugf("%s: sending restart message to %s (%d consecutive restarts)", mc.chid, p, restartCount)
```
aarshkshah1992, line 471: @dirkmc We need to log the fact that we saw a disconnect and that we've successfully sent a Restart at the Info level. Do we do that anywhere ?

dirkmc: Yes on line 451: https://github.com/filecoin-project/go-data-transfer/pull/241/files/e19701052b4d904efe3efb9d7ec1d86545e133af#diff-20982191cb1617ba3ed900d5e26a716e2682ff63cfc53791f3c4feb34927e1c8R451

channelmonitor/channelmonitor.go
```diff
@@ -416,14 +441,14 @@ func (mc *monitoredChannel) doRestartChannel() error {
...
-	log.Infof("%s: restarting (%d consecutive restarts)", mc.chid, restartCount)
+	log.Debugf("%s: restarting (%d consecutive restarts)", mc.chid, restartCount)
 	err := mc.sendRestartMessage(restartCount)
 	if err != nil {
 		log.Warnf("%s: restart failed, trying again: %s", mc.chid, err)
 		// If the restart message could not be sent, try again
 		return mc.doRestartChannel()
 	}
-	log.Infof("%s: restart completed successfully", mc.chid)
+	log.Debugf("%s: restart completed successfully", mc.chid)
```
aarshkshah1992, line 451: @dirkmc This should be Info. It'll only be once per successful debounced restart and will help us debug better.

channelmonitor/channelmonitor.go
```diff
@@ -187,8 +187,33 @@ func newMonitoredChannel(
...
+
+	// "debounce" calls to restart channel, ie if there are multiple calls in a
+	// short space of time, only send a message to restart the channel once
+	var lk sync.Mutex
+	var lastErr error
 	debouncer := debounce.New(cfg.RestartDebounce)
-	mpc.restartChannelDebounced = func() { debouncer(mpc.restartChannel) }
+	mpc.restartChannelDebounced = func(err error) {
+		// Log the error at debug level
+		log.Debug(err.Error())
```
Stebalien, line 198: FYI:

```go
log.Debugw("what does this mean", "error", err)
```

Otherwise:

1. We pay for the overhead of calling `Error`, even if we don't have debug logging enabled.
2. The user will see log messages like "stream reset" with no context.

PR #243, opened 2021-08-24
Limit Push Graphsync requests

hannahhoward:
# Goals

- A provider should be able to limit simultaneous requests for both STORAGE and RETRIEVAL deals
- A provider should be protected from a very active client who sends lots and lots of storage deals

# Implementation

- When Graphsync was written, it was assumed that requests are always initiated by a real human user, meaning rate limiting outgoing Graphsync requests is both unneccesary and actually counterproductive.
- Data transfer's Push mode complicates this assumption. A data transfer push request, while validated with a validator, is initiated by a remote peer, and triggers a Graphsync request. In effect, it's an RPC call to initiate a graphsync request
- As such, it's vulnerable to DDOS attacks from what appears to be the outgoing request side of graphsync
- This PR employs the same mechanism we've used elsewhere -- the peertaskqueue with a set number of simultaneous workers -- to rate limit graphsync requests triggered from remote peers sending Push requests through go-data-transfer
- In keeping with data transfers general architecture of using the Transport layer to adapt a protocol library to the needs of data transfers control layers,  the queue of requests is implemented at the Graphsync transport layer, on the theory that rate limiting is going to be different depending on the protocol
- Concretely, we introduce a RequestQueue that manages request with the PeerTaskQueue -- this both limits the number of simultaneous requests for Graphsync AND load balances among peers prioritizing everyone getting 1 request for a single demanding peers gets lots of requests served
- At the graphsync transport layer, when opening a channel, if the Graphsync request is initiated by the other peer (not us), we put it in the request queue, and it executes when there is an available worker in the queue
- We still execute graphsync requests initiated by our peer immediately -- we wouldn't want to rate limit say initiating 100 retrievals with 100 different providers.
- When deferring the requests, we deferring requests, we dispatch a "TransferQueued" event similar to how we do for the side receiving an incoming Graphsync request.
- There's a test to demonstrate the rate limiting, and the functioning of the peertaskqueue at this point is generally well proven through it's use in Bitswap and go-graphsync

# For discussion

- Should we dispatch a TransferredRequestQueued for outgoing graphsync requests initiated locally? This would close the loop so that every single data transfer from every single side got this event
- We do not put restarts in the queue for now -- the assumption is we want these to get run immediately. As currently written this does free the queue when the original request gets cancelled. Can't tell if this is a problem.
- There are potential additional races introduced here due to more go-routines and asynchronous behavior -- it is probably neccessary to test this heavily before promoting all the way to Lotus master.


transport/graphsync/graphsync.go
```diff
@@ -149,19 +171,46 @@ func (t *Transport) OpenChannel(ctx context.Context,
 		exts = append(exts, doNotSendExt)
 	}
 
-	// Start tracking the data-transfer channel
 	ch := t.trackDTChannel(channelID)
 
+	// if we initiated this data transfer or the channel is open, run it immediately, otherwise
+	// put it in the request queue to rate limit DDOS attacks for push transfers
+	if channelID.Initiator == t.peerID || ch.isOpen {
```
aarshkshah1992, line 178: @hannahhoward Why do we need the `ch.isOpen` condition here  ?

@dirkmc What does `ch.Open` mean semantically for the data-transfer ?

transport/graphsync/graphsync.go
```diff
@@ -906,6 +956,13 @@ func (c *dtChannel) open(ctx context.Context, chid datatransfer.ChannelID, dataS
 		// Mark the channel as open and save the Graphsync request key
 		c.isOpen = true
 		c.gsKey = &gsKey
+		if c.startPaused {
+			err := c.gs.PauseRequest(c.gsKey.requestID)
```
aarshkshah1992, line 960: @hannahhoward - What is the need for this code blob ?

transport/graphsync/graphsync.go
```diff
@@ -980,6 +1037,11 @@ func (c *dtChannel) pause() error {
 	c.lk.Lock()
 	defer c.lk.Unlock()
 
+	if c.gsKey == nil {
```
aarshkshah1992, line 1040: So, this is to support the case where `PauseChannel` is called before we've started the graphsync transfer for a channel that's already opened, right ?

aarshkshah1992 (commented):
Mostly comments.

impl/restart_integration_test.go
```diff
@@ -214,7 +214,7 @@ func TestRestartPush(t *testing.T) {
 
 			// WAIT FOR DATA TRANSFER TO FINISH -> SHOULD WORK NOW
 			// we should get 2 completes
-			_, _, err = waitF(10*time.Second, 2)
+			_, _, err = waitF(100000*time.Second, 2)
```
dirkmc, line 217: Looks like maybe a change got added while debugging?

transport/graphsync/graphsync.go
dirkmc: Once graphsync has called the outgoing requests hook, the channel is open (`isOpen` is true).

I think the idea here is that if the transfer is restarted, data-transfer will call `OpenChannel` again. So we don't want to put restarts at the back of the queue (`OpenChannel` will cancel the existing request and start a new one in its place)

dirkmc: We may need to do a bit more refactoring here.
In the case where a transfer is restarted while the request is still in the queue we probably want to just replace the original request.

transport/graphsync/graphsync.go
```diff
@@ -149,19 +171,46 @@ func (t *Transport) OpenChannel(ctx context.Context,
...
-	// Start tracking the data-transfer channel
 	ch := t.trackDTChannel(channelID)
 
+	// if we initiated this data transfer or the channel is open, run it immediately, otherwise
+	// put it in the request queue to rate limit DDOS attacks for push transfers
+	if channelID.Initiator == t.peerID || ch.isOpen {
+		// Start tracking the data-transfer channel
+
+		// Open a graphsync request to the remote peer
+		req, err := ch.open(ctx, channelID, dataSender, root, stor, doNotSendCids, exts)
```
dirkmc, line 182: Suggest passing the `ch` object, rather than passing `channelID` here and then calling `getDTChannel(channelID)` in `runDeferredRequest`

transport/graphsync/requestqueue/requestqueue.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (orq *RequestQueue) Start(ctx context.Context) {
+	for i := uint64(0); i < orq.maxInProcessRequests; i++ {
+		go orq.processRequestWorker(ctx)
+	}
+}
+
+func (orq *RequestQueue) processRequestWorker(ctx context.Context) {
```
dirkmc, line 48: In this model there are many threads that all get "woken up" each time a task is added to the queue, and then all try to pop tasks from the queue, but most of the time all but one will get an empty task list.

I wonder if it would be possible to change the model so that there are initially zero go routines running. When a task is added to the queue a new go routine is started to process it. Each time a new go routine is needed it's started, up to some maximum number of go routines. As tasks complete, if the go routine is no longer needed it stops, all the way down to zero go routines.

This would allow us to use the minimum number of required go routines, and we would no longer need the `Start()` method.

I'm not sure if it's possible, but it would be nice if it is possible.

transport/graphsync/graphsync.go
aarshkshah1992: @dirkmc 

You are right. So, if a queued transfer gets restarted, this PR will add a duplicate request to the queue, right ?

dirkmc: Yes, for push requests I think it will add a duplicate request to the queue.

PR #254, opened 2021-09-03
Fix parallel transfers between same two peers

hannahhoward:
# Goals

HAT TIP @tchardin !!! You may have caught a fairly critical bug.

Your bug report and submitted test revealed a serious issue in the go-graphsync responder implementation.

# Implementation

This PR contains some fixes to @tchardin 's test to insure it properly catches the issue in question, and then an update to go-graphsync (https://github.com/ipfs/go-graphsync/pull/211) that fixes the issue.


dirkmc (approved):
LGTM 👍 

Thanks @tchardin 🙌

PR #257, opened 2021-10-01
Use do-not-send-first-blocks extension for restarts

### LINKED ISSUE #256: When a graphsync transfer restarts, send "do-not-send-first-blocks" message
When a graphsync transfer restarts (because the connection goes down, or because the node is bounced) the party receiving the data makes the same request, and uses an extension to indicate what data it has already received.

Currently the extension contains a list of all CIDs that the receiver already received. However some DAGs can contain so many blocks that the CID list becomes prohibitively long.

To address this problem a "do-not-send-first-blocks" extension will be added to graphsync. The extension allows the receiver to indicate the number of blocks it has received. Blocks are sent in a consistent order, so the sender can use the number of blocks received as an offset into the ordered list of blocks to be sent.

go-data-transfer needs to be updated to use this extension instead of "do-not-send-cids". This will require a protocol version bump.
go-data-transfer should continue to use the "do-not-send-cids" extension with peers using older versions of the protocol.

#### Proposed Implementation

Add method to `DataTransferNetwork` interface:
```
// Get the go-data-transfer protocol that the peer supports, connecting to the peer if necessary
Protocol(peer.ID) protocol.ID
```

Modify `Transport.OpenChannel()` interface to
- take a `ChannelState` instead of a `doNotSendCids` parameter
  Note that `ChannelState` has `ReceivedCids()` and `ReceivedCidsLen()` methods

In the graphsync implementation of `Transport`:
- Modify `NewTransport()` to take an interface that exposes `DataTransferNetwork`'s `Protocol(peer.ID) protocol.ID` method
- Modify `OpenChannel()`:
  if `ReceivedCidsLen()` > 0
    - switch on the peer's protocol ID
    - if the protocol is high enough, add a "do-not-send-first-blocks" extension
    - otherwise add the old "do-not-send-cids" extension

dirkmc:
Fixes #256


hannahhoward (commented):
Can you explain the rationale for the blocksReceivedIndex? Why not just add a value to the main data transfer state? It's just a number -- the only reason we store received cids list seperately is they are very large

transport.go
```diff
@@ -93,13 +92,15 @@ type Transport interface {
...
-	OpenChannel(ctx context.Context,
+	OpenChannel(
+		ctx context.Context,
 		dataSender peer.ID,
 		channelID ChannelID,
 		root ipld.Link,
 		stor ipld.Node,
-		doNotSendCids []cid.Cid,
-		msg Message) error
+		channel ChannelState,
```
hannahhoward, line 101: sidebar +++1 for this change generally. I've never liked the specificity of doNodeSendCids. Technically, we can probably drop a bunch of other parameters if we do this (I believe root/selector/sender/channelID can all be extracted from channel state)

hannahhoward: I see that wouldn't quite work for the code here just yet (you're counting on channel == nil being evidence of a not-restart), but a great thing for future refactor!

transport/graphsync/graphsync.go
```diff
@@ -164,6 +166,69 @@ func (t *Transport) OpenChannel(ctx context.Context,
...
 
+// Get the extension data for sending a Restart message, depending on the
+// protocol version of the peer
+func (t *Transport) getRestartExtension(ctx context.Context, p peer.ID, channel datatransfer.ChannelState) ([]graphsync.ExtensionData, error) {
+	if channel == nil {
+		return nil, nil
+	}
+
+	// Get the peer's protocol version
+	protocol, err := t.peerProtocol.Protocol(ctx, p)
```
hannahhoward, line 177: So I want to point out something that's a potential reason to augment the Graphsync extension as well: in the case of a pull request, you may still not have this cached even for a restart, as the only communication may be over graphsync. I don't know whether this matters a lot, but the idea would be to maintain your protocol list outside the network implementation, so you can update it for an incoming GS request/response as well as well. The only penalty is a new libp2p message I guess, so not the worst thing but still.

hannahhoward (changes requested):
See previous comment about blocks received index.

Comment about tracking peers outside the go-graphsync extension and tracking peers outside the network impl is NON-BLOCKING.

transport/graphsync/graphsync.go
dirkmc: I see what you mean, that may save us an extra round trip.

If I understand correctly, if we add another extension protocol, then whenever we send a message, we also include an encoded payload for that extension. Is that correct? So for example if we have an extension for go-data-transfer-v1.2, the message will look like:
```
<message data>
extensions:
  fil/data-transfer/1.2: <encoded data>
  fil/data-transfer/1.1: <encoded data>
  fil/data-transfer:     <encoded data>
```

Maybe easier to explain with a screenshot from the debugger:

Screen Shot 2021-10-01 at 6 46 24 PM
:

dirkmc:
> Can you explain the rationale for the blocksReceivedIndex

You're right it makes more sense to store it on the state. I've pushed a commit to do that.

hannahhoward (approved):

transport/graphsync/graphsync.go
hannahhoward: yes it's an additional copy of the exact same mesage... so arguably this ones a bit a bummer, especially since there's no difference in the message format.... it's literally just the extension name.

BTW we can probably deprecate 1_0 at this point -- hasn't existed since pre-mainnet.

dirkmc: Also Aarsh pointed out that for a pull request it's the data receiver who *sends* the extension, so the data receiver is not going to know the protocol of the data sender.
So I think it makes most sense not to use an additional extension for now.

network/libp2p_impl.go
```diff
@@ -289,8 +316,52 @@ func (dtnet *libp2pDataTransferNetwork) msgToStream(ctx context.Context, s netwo
...
 	}
 
-	if err := s.SetWriteDeadline(time.Time{}); err != nil {
-		log.Warnf("error resetting deadline: %s", err)
-	}
 	return nil
 }
+
+func (impl *libp2pDataTransferNetwork) Protocol(ctx context.Context, id peer.ID) (protocol.ID, error) {
+	// Check the cache for the peer's protocol version
```
aarshkshah1992, line 323: @dirkmc You don't need this cache as libp2p maintains it's own cache. You can use `impl.host.Peerstore().FirstSupportedProtocol()` for the cache lookup and then go do the open stream thing it it doesen't have anything for you.

aarshkshah1992 (commented):
@dirkmc First set of comments .

channels/channels.go
```diff
@@ -241,8 +241,27 @@ func (c *Channels) DataQueued(chid datatransfer.ChannelID, k cid.Cid, delta uint
 }
 
 // Returns true if this is the first time the block has been received
-func (c *Channels) DataReceived(chid datatransfer.ChannelID, k cid.Cid, delta uint64) (bool, error) {
-	return c.fireProgressEvent(chid, datatransfer.DataReceived, datatransfer.DataReceivedProgress, k, delta)
+func (c *Channels) DataReceived(chid datatransfer.ChannelID, k cid.Cid, delta uint64, index int64) (bool, error) {
```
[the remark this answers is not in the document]
aarshkshah1992: After a restart -> GS will fire these notifications for blocks we have already received, right ? So, we'll re-update the offset for each block all the way to the last received block and therefore the latest offset, right ?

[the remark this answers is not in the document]
dirkmc: Yes 👍

PR #259, opened 2021-10-04
remove libp2p protocol cache

dirkmc:
libp2p has its own protocol cache, so there's no need to keep on external to libp2p

### DIFF
--- a/network/libp2p_impl.go
+++ b/network/libp2p_impl.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"sync"
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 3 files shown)


network/libp2p_impl.go
```diff
@@ -341,27 +327,12 @@ func (impl *libp2pDataTransferNetwork) Protocol(ctx context.Context, id peer.ID)
...
-	defer impl.pplk.Unlock()
-
-	delete(impl.peerProtocols, p)
-}
-
-// Clear the peer protocol version cache for the peer when the peer disconnects
-func (impl *libp2pDataTransferNetwork) Disconnected(n network.Network, conn network.Conn) {
-	impl.clearPeerProtocol(conn.RemotePeer())
+	// Keep a string version of the protocols for performance reasons
+	impl.dtProtocolStrings = make([]string, 0, len(impl.dtProtocols))
```
aarshkshah1992, line 334: @dirkmc Why do we need this now ?

dirkmc: Because `FirstSupportedProtocol(id, protos...)` takes an array of strings (not an array of `protocol.ID`)

aarshkshah1992 (commented):
Approved.

PR #283, opened 2021-11-19
update to context datastores

whyrusleeping:


testutil/fakegraphsync.go
```diff
@@ -422,6 +422,11 @@ func (fgs *FakeGraphSync) Stats() graphsync.Stats {
 	return graphsync.Stats{}
 }
 
+func (fgs *FakeGraphSync) RegisterOutgoingRequestProcessingListener(graphsync.OnOutgoingRequestProcessingListener) graphsync.UnregisterHookFunc {
```
arajasek, line 425: where this came from why this exist

arajasek (approved):

arajasek:
Updated to latest go-ds-versioning, but test seems to be flaky failing (passes locally)

testutil/fakegraphsync.go
hannahhoward: yea that's a mock GraphExchange interface that needs a new method cause Graphsync updated and added it.. I'm moving away from putting methods on GraphExchange and towards only implementing them on the main implementation to simplify this. But later.

hannahhoward:
@arajasek go-data-transfer I believe still suffers from unreliable CI -- I haven't been in it in a bit but @dirkmc can confirm.

hannahhoward (approved):
LGTM

PR #287, opened 2021-12-07
Improve GraphSync Transport Diagnostics

hannahhoward:
# Introduction

This will be the first of a few PRs aimed at improving diagnostics and tracing in go-data-transfer. A big caveat: our current situation is not ideal, and we plan to do big refactors in go-data-transfer soon. But that does not alleviate the need for diagnostic and tracing information in the existing code right now. As a result, these PRs will rarely change functionality and where new functionality is added, it should generally not be taken as permanent designs - we may need to do a few things in less than ideal ways while we lay the groundwork for bigger changes.

# Goals

This PR adds two things. First, it makes some minor changes to GraphSync transport logs.

Second, it adds a method for introspecting the current Graphsync transport for a given peer we're doing transfers with. Specifically it exposes mapping graphsync request IDs to data transfer channel IDs. Ultimately, where we are headed is a diagnostic endpoint in Estuary aimed to give in depth information about the current state of transfers with a peer and their corresponding graphsync states. Again, all of this state is headed for a big refactor but we need to be able to see it now to diagnose and solve bugs.

Asking for a bit of grace in merging some slightly less than ideal code changes in order to move towards a long term goal with iterative improvements.


rvagg (approved):
rubber stamp lgtm, some of this is opaque to me but looks good on the surface

transport/graphsync/graphsync.go
```diff
@@ -404,6 +404,43 @@ func (t *Transport) UseStore(channelID datatransfer.ChannelID, lsys ipld.LinkSys
 	return ch.useStore(lsys)
 }
 
+type ChannelGraphsyncRequests struct {
+	Current graphsync.RequestID
+	Other   []graphsync.RequestID
```
dirkmc, line 409: Could you please add some comments explaining what `Current` and `Other` are

transport/graphsync/graphsync.go
```diff
@@ -404,6 +404,43 @@ func (t *Transport) UseStore(channelID datatransfer.ChannelID, lsys ipld.LinkSys
...
+				collection = receiving
+			}
+			channelGraphsyncRequests := collection[chid]
+			if t.dtChannels[chid] != nil && t.dtChannels[chid].gsKey != nil && (*t.dtChannels[chid].gsKey) == k {
+				channelGraphsyncRequests.Current = k.requestID
+			} else {
+				channelGraphsyncRequests.Other = append(channelGraphsyncRequests.Other, k.requestID)
+			}
+			collection[chid] = channelGraphsyncRequests
+		}
```
dirkmc, line 436: Please add some comments explaining what this code is doing

dirkmc (commented):
I can follow the logic, but I'm not sure why we need the `Other` field.
Please add some comments to clarify why each field is needed.

hannahhoward:
@dirkmc I will add said comments. What I am seeing is that as transfers get restarted multiple times, they acquire multiple graphsync request IDs. I am concerned about some of these request IDs "hanging around" in graphsync's queue, when they shouldn't. I'm trying to track down what's happening, so I'm just exposing these these records.

hannahhoward:
note that as I read the code, gsKeyToChannelIDMap doesn't wipe requests until the whole transfer closes.

dirkmc:
Thanks, makes sense and the comments look great 👍

PR #290, opened 2022-01-06
Block spans

hannahhoward:
# Goals

Add traces for block progress events to see how much time we're spending on these.

# Implementation

Add tracing spans to DataReceived, DataSent, and DataQueued


rvagg (approved):
just naming convention thing again, spans with `-` in their names are going to stand out in the stack like there's something special about them, or the other way around, nbd

rvagg (approved):

PR #291, opened 2022-01-13
Accept missing blocks without failing data transfer

hannahhoward:
# Goals

A data transfer should be able to transfer all data a remote peer has in a DAG, even if it's missing parts of the DAG. Graphsync already supports this at the protocol level, but we've been failing data transfers any time a block is missing.

Now, instead of failing, we record that the sender was missing a CID and move on. This enables us to store data on the Filecoin network where DAGs do not always go all the way to the leaf without having to support super complex selectors.

# Implementation

- add new event CIDMissing
- add missing CIDs array member on the transfer state. Note: I am not so concerned about just having this live directly on data transfer, as it happens rarely, and moreover, keep in mind if a CID is missing, the traversal in graphsync skips over it, so we won't ever learn about its child CIDs. I don't expect therefore this list will be super long
- add new event hook on the transport callback for missing CIDs
- read errors in the graphsync error stream and produce calls to the missing link callback on events.

# For Discussion

- Limitations:
   - The sending side does not learn about missing CIDs. Graphsync does not expose this data.
   - Currently the final status on both sides is simply "Completed". Do we need to do the work to create a PartiallyComplete final status? The upside is more obvious behavior. The downside is code breakage upstream. Also, it'd be hard to put both sides in Partially Complete unless Graphsync exposes missing CIDs to the sending side, so we'd have to accept differing final statuses for now or do the work to make Graphsync expose missing CIDs to the sender now.
   
- Blockers:
Merge https://github.com/ipfs/go-graphsync/pull/331


channels/channels_fsm.go
```diff
@@ -103,6 +104,24 @@ var ChannelEvents = fsm.Events{
 			chst.AddLog("")
 			return nil
 		}),
+
+	fsm.Event(datatransfer.CIDMissing).FromMany(transferringStates...).ToJustRecord().
+		Action(func(chst *internal.ChannelState, missing cid.Cid) error {
+			// TODO: find a more efficient way to do this
```
dirkmc, line 110: Would it make sense to store the missing cids as a map[cid.Cid]struct{} instead?

transport/graphsync/graphsync.go
```diff
@@ -238,6 +238,11 @@ func (t *Transport) consumeResponses(req *gsReq) error {
 	log.Infof("channel %s: finished consuming graphsync response channel", req.channelID)
 
 	for err := range req.errChan {
+		var remoteMissingBlockErr graphsync.RemoteMissingBlockErr
+		if errors.As(err, &remoteMissingBlockErr) {
+			t.events.OnLinkMissing(req.channelID, remoteMissingBlockErr.Link)
+			continue
+		}
```
dirkmc, line 245: Suggest adding some comments here - I get what it's doing from the explanation in the PR, but someone else reading this code later may not make the connection

dirkmc (commented):
The code LGTM 👍 

Conceptually I'd be a little wary of saying a request has completed when some blocks are missing. For the node downloading the data this could lead to confusing results.

I think the downloader should be able to differentiate between
- all blocks downloaded
- some blocks not downloaded because the transfer was interrupted part way through
- traversal completed but some blocks not downloaded because the sender doesn't have them

channels/channels_fsm.go
dirkmc: Ok 👍

Issue #301, opened 2022-02-18
Graphsync 2.0 refactors: Bind node part 1 - message


hannahhoward:
# Goals

#300 integrates the Graphsync 2.0 branch and passes tests.

However, it uses a pretty serious hack to make the GraphSync extension work. Graphsync extensions now MUST be a [go-ipld-prime datamodel.Node](https://github.com/ipld/go-ipld-prime/blob/master/datamodel/node.go#L58). In order to make this work with data transfer messages, which are currently a go struct with cbor-gen encoding/decoding, we employ a pretty inefficient trick to coerce the data into a datamodel.Node -- we simply serialize to CBOR and deserialize using basicnode.Any.

A better strategy is to use [bindnode](https://github.com/ipld/go-ipld-prime/tree/master/node/bindnode) -- this is just more work and may push on features of bindnode that are still experimental.

So, here's what is needed:

- define a schema for data transfer messages - [transferMessage](https://github.com/filecoin-project/go-data-transfer/blob/master/message/message1_1/transfer_message.go), [transferRequest](https://github.com/filecoin-project/go-data-transfer/blob/master/message/message1_1/transfer_request.go), [transferResponse](https://github.com/filecoin-project/go-data-transfer/blob/master/message/message1_1/transfer_response.go) -- they may may need renames to play nicely with schema (shouldn't affect encoding).
   - my read of pointer type encoding in cbor-gen is the fields should be nullable but not optional
   - use map representation
- change any cbg.Deferred types to datamodel.Node and in the schema use ANY.
- handle encoding of vouchers to datamodel.Node -- here again we will employ a hack for now, similar to the one we use with messages currently -- I think we'll need to make a new method in [encoding](https://github.com/filecoin-project/go-data-transfer/blob/master/encoding/encoding.go) -- `EncodeToNode(encoding.Encodable) (datamodel.Node, error)` -- and a new method for the `Decoder` interface -- `DecodeFromNode(datamodel.Node) (encoding.Encodable, error)`
- update [message constructors](https://github.com/filecoin-project/go-data-transfer/blob/master/message/message1_1/message.go)
- update ToIPLD function -- I believe this simply becomes bindnode.Wrap with schema, followed by calling Representation() (so it serializes properly in Graphsync)
- update FromIPLD function -- I believe this looks like bindnode.Prototoype and then calling Representation().NewBuidler().AssignNode()then bindnode.Unwrap on the built node.

Potential issues:
[...]

PR #304, opened 2022-02-22
Graphsync v2.0: bindnode conversions

rvagg:
Continues #300

~~Currently stuck on https://github.com/ipld/go-ipld-prime/issues/365 in `FromIPLD()`.~~

From the list in #301, we have so far, some of which was already done in `feat/graphsync-20-minimal`:

* [x] define a schema for data transfer messages
* [x] change any cbg.Deferred types to datamodel.Node and in the schema use ANY.
* [x] handle encoding of vouchers to datamodel.Node -- _we're just doing the round-trip through dag-cbor to achieve this, nothing fancy about voucher handling yet__
* [x] update message constructors
* [x] update ToIPLD function
* [x] update FromIPLD function

And comments on the list of potential problems:

* we have a datatransfer.ChannelID in one of the message types  -- _the schema for this has just been rolled up into the messages schema, it's pretty simple and maps cleanly using bindnode. There were [challenges with the tuple representation](https://github.com/ipld/go-ipld-prime/issues/367) that had to get fixed along the way, but it works now._
* unsigned int usage -- _was [fixed](https://github.com/ipld/go-ipld-prime/pull/363) and works fine_
* we need to maintain byte to byte CBOR-gen serialization compatibility -- _this current branch includes both the old and new message formats and both contain serialization tests from the other one, so the old messages contain tests that check that binary from ipld-prime can be decoded, and the tests in the new messages contain tests that check that binary from cbor-gen can be decoded. They pass. (I also did some work on the [cborg](https://github.com/rvagg/cborg) CLI to poke at the internal CBOR to do some checking on the differences and confirmed that it's just the map key ordering, and no codec we have cares about map ordering on decode so we're fine)._
* does Representation().NewBuilder.AssignNode() actually work -- _no it didn't! but now it does, main fix was https://github.com/ipld/go-ipld-prime/issues/365_ and the tuple one was another fix needed to make it work._

### DIFF
--- a/encoding/encoding.go
+++ b/encoding/encoding.go
@@ -7,6 +7,9 @@ import (
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/schema"
 	cborgen "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 )
@@ -25,8 +28,11 @@ func Encode(value Encodable) ([]byte, error) {
 		}
 		return buf.Bytes(), nil
 	}
-	if ipldEncodable, ok := value.(ipld.Node); ok {
-		buf := new(bytes.Buffer)
+	if ipldEncodable, ok := value.(datamodel.Node); ok {
+		if tn, ok := ipldEncodable.(schema.TypedNode); ok {
+			ipldEncodable = tn.Representation()
+		}
+		buf := &bytes.Buffer{}
 		err := dagcbor.Encode(ipldEncodable, buf)
 		if err != nil {
 			return nil, err
@@ -36,17 +42,30 @@ func Encode(value Encodable) ([]byte, error) {
 	return cbor.DumpObject(value)
 }
 
+func EncodeToNode(encodable Encodable) (datamodel.Node, error) {
+	byts, err := Encode(encodable)
+	if err != nil {
+		return nil, err
+	}
+	na := basicnode.Prototype.Any.NewBuilder()
+	if err := dagcbor.Decode(na, bytes.NewReader(byts)); err != nil {
+		return nil, err
+	}
+	return na.Build(), nil
+}
+
 // Decoder is CBOR decoder for a given encodable type
 type Decoder interface {
 	DecodeFromCbor([]byte) (Encodable, error)
+	DecodeFromNode(datamodel.Node) (Encodable, error)
 }
 
 // NewDecoder creates a new Decoder that will decode into new instances of the given
 // object type. It will use the decoding that is optimal for that type
 // It returns error if it's not possible to setup a decoder for this type
 func NewDecoder(decodeType Encodable) (Decoder, error) {
-	// check if type is ipld.Node, if so, just use style
-	if ipldDecodable, ok := decodeType.(ipld.Node); ok {
+	// check if type is datamodel.Node, if so, just use style
+	if ipldDecodable, ok := decodeType.(datamodel.Node); ok {
 		return &ipldDecoder{ipldDecodable.Prototype()}, nil
 	}
 	// check if type is a pointer, as we need that to make new copies
@@ -86,6 +105,14 @@ func (decoder *ipldDecoder) DecodeFromCbor(encoded []byte) (Encodable, error) {
 	return builder.Build(), nil
 }
 
+func (decoder *ipldDecoder) DecodeFromNode(node datamodel.Node) (Encodable, error) {
+	builder := decoder.style.NewBuilder()
+	if err := builder.AssignNode(node); err != nil {
+		return nil, err
+	}
+	return builder.Build(), nil
+}
+
 type cbgDecoder struct {
 	cbgType reflect.Type
 }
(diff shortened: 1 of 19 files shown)


mvdan:
ipld-prime fix here :) https://github.com/ipld/go-ipld-prime/pull/366

rvagg:
Fix is 👌 , next problem is https://github.com/ipld/go-ipld-prime/issues/367

rvagg:
Marked as ready-to-review and updated the original comment to tick off items from #301. I think we're done here and tests are passing minus some flakies.

The only major thing we could do to this branch is remove the old `messages1_1` package and replace it with the `messages1_1prime` package, but all the dependent code is currently ignoring the old one and using the new one anyway, and it's been handy to keep it in place to test compatibility. Maybe this is something that could be done prior to the merging of `feat/graphsync-20-minimal` in #300, because I might do a similar thing with vouchers.

encoding/encoding.go
```diff
@@ -36,17 +42,30 @@ func Encode(value Encodable) ([]byte, error) {
...
 
 // NewDecoder creates a new Decoder that will decode into new instances of the given
 // object type. It will use the decoding that is optimal for that type
 // It returns error if it's not possible to setup a decoder for this type
 func NewDecoder(decodeType Encodable) (Decoder, error) {
-	// check if type is ipld.Node, if so, just use style
-	if ipldDecodable, ok := decodeType.(ipld.Node); ok {
+	// check if type is datamodel.Node, if so, just use style
+	if ipldDecodable, ok := decodeType.(datamodel.Node); ok {
 		return &ipldDecoder{ipldDecodable.Prototype()}, nil
```
hannahhoward, line 69: should we be checking for schema.TypedPrototype here and using Prototype().Representation()

hannahhoward: probably doesn't affect this PR but I expect it affects the next one?

message/message1_1/transfer_message.go
```diff
@@ -4,31 +4,32 @@ import (
...
+
+	datatransfer "github.com/filecoin-project/go-data-transfer"
 )
 
-//go:generate cbor-gen-for --map-encoding transferMessage1_1
+//go:generate cbor-gen-for --map-encoding TransferMessage1_1
 
 // transferMessage1_1 is the transfer message for the 1.1 Data Transfer Protocol.
-type transferMessage1_1 struct {
+type TransferMessage1_1 struct {
```
hannahhoward, line 17: Question: if these are all the old types, why are we renaming them?

hannahhoward: it just makes the change set larger

hannahhoward: also, since they exist ONLY for a compatibility test, I wonder if the directory structure should be simply

`message1_1` - Prime nodes
`message1_1/cbor_gen_legacy_test` (or something like that) - this directory?

transport/graphsync/graphsync_test.go
```diff
@@ -1338,21 +1338,31 @@ func assertDecodesToMessage(t *testing.T, data datamodel.Node, expected datatran
 func assertHasOutgoingMessage(t *testing.T, extensions []graphsync.ExtensionData, expected datatransfer.Message) {
 	nd, err := expected.ToIPLD()
 	require.NoError(t, err)
-	expectedExt := graphsync.ExtensionData{
-		Name: extension.ExtensionDataTransfer1_1,
-		Data: nd,
+	found := false
+	for _, e := range extensions {
+		if e.Name == extension.ExtensionDataTransfer1_1 {
+			require.True(t, ipld.DeepEqual(nd, e.Data), "data matches")
```
hannahhoward, line 1344: This is cause e.Data is a generic basicnode.Any node while nd is a representation of the message node format?

hannahhoward: also I'm glad ipld.DeepEqual exists :)

message/message1_1prime/transfer_message.go
```diff
@@ -0,0 +1,44 @@
...
+// TransferID returns the TransferID of this message
+func (tm *TransferMessage1_1) TransferID() datatransfer.TransferID {
+	if tm.IsRequest {
+		return tm.Request.TransferID()
+	}
+	return tm.Response.TransferID()
+}
+
+func (tm *TransferMessage1_1) toIPLD() schema.TypedNode {
+	return bindnode.Wrap(&tm, Prototype.TransferMessage.Type())
```
hannahhoward, line 33: just noting you have about 6x places return value here is ultimately used and all of them call .Representation() immediately. I wonder if you  just return the representation node and save yourself the trouble.

hannahhoward (approved):
LGTM, go ahead and merge to the branch if you like.

All my comments are suggestions and you're welcome to fix before or after merge. My strongest suggestion is around the directory structure.

transport/graphsync/graphsync_test.go
rvagg: Yeah, I think this is because the data doesn't do a full round-trip through binary so you end up with a typed version which fails a normal match, or something like that. It was just easier to do a DeepEqual than try and make the types perfectly the same. But we could also force a round-trip somewhere in here.

I'm reconsidering some of the serialisation stuff in the next PR, maybe these things will change a little by the time I get back to here.

message/message1_1prime/transfer_message.go
rvagg: Yes .. the original version of my code does have it, but then there was one place I found I wanted it back to a TypedNode.. Again I think some of the serialisation stuff will change so maybe this can go back. At least in the current state of my next PR this method has entirely changed and bindnode logic has been moved elsewhere .. but that's in flux while I work my head through the rest of the system.

message/message1_1/transfer_message.go
rvagg: It was something about getting the tests to pass, which are in a `message1_1test` package and are circular if you remove the `test`. I don't recall why I needed to do this but it ended up being the easiest path to solving it ... In my next PR I've just removed this directory, it became too hard to keep intact because it pulls in the FakeDType stuff which needs to change too.

encoding/encoding.go
rvagg: good Q, but maybe it doesn't even matter once we're dealing with a `datamodel.Node` interface, I think this whole encoding package can go away.

Issue #297, opened 2022-02-04
New Revalidator API


hannahhoward:
# Goals

The current API for revalidation places a heavy burden on the markets / exchange layer tracking progress on a data transfer and inspecting every block sent for decisions about pausing.

We should make it easier for an exchange layer to only track the state relative to actual payments -- namely, how much has been paid, and given the current state of transfer, how much is owed.

# Implementation

The new proposed API is as follows:

```golang

type CheckPoint struct{
    // First point after which to pause and ask for payment
    ByteOffset uint64
    // treat completing request as a checkpoint if it happens before next nextOffset
    PauseOnComplete bool
}

type Revalidator interface {

	// NextCheckPoint specifies the next point at which data transfer should
	// pause transfer
	// return values are:
	// has = are there remaining checkpoints for this request? If not continue to end
        // checkpoint = next checkpoint
	// err = abort request if not nil
	NextCheckPoint(
		channelID datatransfer.ChannelID,
                 bytesSoFar uint64
	) (has bool, checkpoint Checkpoint, err error)

	// CheckpointReached allows an exchange to produce a request for payment
	// after a checkpoint has
	// been passed
	// Note: checkpoint is not exact for some protocols -- the byteOffs may be
	// slightly larger than
	// the one requested
	// return values are:
	// voucher result = information about payment requested, if any
	// err = error
	// - nil = pause request
	// - error = abort this request
	CheckpointReached(
		channelID datatransfer.ChannelID,
		actualByteOffset uint64,
	) (datatransfer.VoucherResult, error)

	// EndOfDataReached allows for a final last request for payment
	// return values are:
	// voucher result = information about payment requested, if any
	// err = error
	// - nil = pause request
	// - error = terminate request with error
	EndOfDataReached(
		channelID datatransfer.ChannelID,
		actualByteOffset uint64,
	) (datatransfer.VoucherResult, error)

	// Revalidate processes a new voucher (likely a payment) intended to resume
	// transfer
	// return values are:
	// voucher result = information about why the payment succeeded or failed
	// resume = should request resume? or stay paused for more revalidation
	// err = error
	// - nil = resume if resume = true
	// - error = abort this request
	Revalidate(
		channelID datatransfer.ChannelID,
		voucher datatransfer.Voucher,
		actualByteOffset uint64,
	) (voucherResult datatransfer.VoucherResult, resume bool, err error)
}
```

[...]

PR #308, opened 2022-03-02
Refactor revalidators

hannahhoward:
# Goals

Implement #297 

# Implementation

The changes here are, to be blunt, not small. I recommend reviewing in chunks:
- The changes to the state machine and new events
- The changes to the message library
- The changes in validation and event processing

The major overall change is to the validation process. Here's our new validation interface:

```golang
// ValidationResult describes the result of validating a voucher
type ValidationResult struct {
	// Accepted indicates where the request was accepted if a request is not
	// accepted, the request fails. This is true for revalidation as well
	Accepted bool
	// VoucherResult provides information to the other party about what happened
	// with the voucher
	VoucherResult
	// LeaveRequestPaused indicates whether the request should stay paused
	// even if the request was accepted
	LeaveRequestPaused bool
	// DataLimit specifies how much data this voucher is good for. When the amount
	// amount data specified is reached (or shortly after), the request will pause
	// pending revalidation. 0 indicates no limit.
	DataLimit uint64
	// RequiresFinalization indicates at the end of the transfer, the channel should
	// be left open for a final settlement
	RequiresFinalization bool
}

// RequestValidator is an interface implemented by the client of the
// data transfer module to validate requests
type RequestValidator interface {
	// ValidatePush validates a push request received from the peer that will send data
	// -- All information about the validation operation is contained in ValidationResult,
	// including if it was rejected. Information about why a rejection occurred is embedded
	// in the VoucherResult.
	// -- error indicates something went wrong with the actual process of trying to validate
	ValidatePush(
		chid ChannelID,
		sender peer.ID,
		voucher Voucher,
		baseCid cid.Cid,
		selector ipld.Node) (ValidationResult, error)
	// ValidatePull validates a pull request received from the peer that will receive data
	// -- All information about the validation operation is contained in ValidationResult,
	// including if it was rejected. Information about why a rejection occurred should be embedded
	// in the VoucherResult.
	// -- error indicates something went wrong with the actual process of trying to validate
	ValidatePull(
		chid ChannelID,
		receiver peer.ID,
		voucher Voucher,
		baseCid cid.Cid,
		selector ipld.Node) (ValidationResult, error)

	// Revalidate revalidates a request with a new voucher
	// -- All information about the validation operation is contained in ValidationResult,
	// including if it was rejected. Information about why a rejection occurred should be embedded
	// in the VoucherResult.
	// -- error indicates something went wrong with the actual process of trying to validate
	Revalidate(channelID ChannelID, channel ChannelState) (ValidationResult, error)
}
```

What changed:

- Revalidators went away entirely -- every voucher validator should be able to revalidate from the channel state
- The isRestart param went away -- a restart triggers a revalidation
- A validation now produces a LOT more information
   - First, the error parameter in the validation was way overloaded
      - it groups "errors" i.e. something went wrong with my server when I validated with "rejections" -- i.e. your voucher has a problem and I'm not accepting your request (you could think of this as analogous to 500 vs 400 response codes)
         - now an error is only the "something went wrong with my server"
         - the Accepted parameter of ValidationResult now identifies whether a rejection occurred
            - if you need to pass information back to the initiator about why things failed, it goes in the VoucherResult. The previous errors were not returned to the caller anyway, and go-fil-markets retrieval code already relies on sending data in the VoucherResult
      - we also use the "error" to send a Pause/Resume command. This made things even more annoying.
         - now a pause / resume is expressed by LeaveRequestPaused
   - DataLimit allows you to tell data transfer ahead of time how much data this voucher is good for. When the limit gets hit, DataTransfer handles pausing for you. This replaces the OnPushDataReceived / OnPullDataSent methods of the revalidators
   - RequireFinalization allows you to tell data transfer ahead of time that you're gonna need one more round of validation at the end. If set true, data transfer will leave the channel open for you to send some vouchers back and forth -- this replaces the OnComplete method in the revalidator.
   
 One design quirk I'm still thinking about:
 - the way you tell a transfer stopped on a datalimit to resume is setting LeaveRequestPaused false. This doesn't take into account the current data limit, and whether you actually changed it. I wonder if we should put this check in, have the unpause happen from the data limit change and going over the current sent/received, and then rename LeaveRequestPaused as an affirmative "Pause the request now" -- as in pause it if we're not paused already, but don't resume.
 
So that's the main thing, but it was a big enough change I had to make a lot of code changes to do it, and that meant I had to work through a lot of code that I wrote (or others wrote) almost a year ago.... and boy it was hard not to do cleanups, especially since I had to comment to work out what it was doing again. (OY) The upside is this means the code has a lot of comments and that should make reviewing easier. Also I moved all the code related to handling request messages to it's own file, cause that has a ton of logic.

Other changes:
- Core logic for data limits in OnDataQueued/OnDataReceived:
   - I essentially copied the strategy used with tracking data progress without going to the state machine, caching information in memory
   - BOY am I ready to throw them state machines out and replace them with something that gives us an in memory copy all the time but that's a refactor for another time
   
- I changed the `IsVoucherResult` method to `IsValidationResult` for clarity -- this removes a weird pairing of `IsVoucherResult` & `EmptyVoucherResult` methods -- `IsVoucherResult` and `EmptyVoucherResult` could be true at the same time, which was confusing

- Added an imperative SendVoucherResult method. Honestly, I'd rather this not exist, but when we update the retrieval code, I want to maintain compatibility, and this will allow us to send the same information from the server we were already sending when the channel pauses

There were also some low hanging refactors that drove me crazy as I added changes -- The ChannelState implementation was holding it's own copy of the channelstate which had to be updated and copied on -- now it just embeds and internalChannelState struct.
 
There were two implementations of mockChannelState that I consolidated.
 
That's most of it. I don't think this can merge until I complete at least a draft implementation of updating the retrieval code in go-fil-markets - to verify that we actually made possible the big simplifications I'm hoping for.

Also I think this is breaking enough that it should release (along with the transport refactors) as a 2.0 package.


channels/channel_state.go
```diff
@@ -98,60 +57,60 @@ func (c channelState) Selector() ipld.Node {
...
 // IsPull returns whether this is a pull request based on who initiated it
 func (c channelState) IsPull() bool {
-	return c.isPull
+	return c.ic.Initiator == c.ic.Recipient
 }
 
 func (c channelState) ChannelID() datatransfer.ChannelID {
-	if c.isPull {
-		return datatransfer.ChannelID{ID: c.transferID, Initiator: c.recipient, Responder: c.sender}
+	if c.ic.Initiator == c.ic.Recipient {
```
dirkmc, line 101:
```suggestion
	if c.IsPull() {
```

channels/channels.go
```diff
@@ -354,6 +373,17 @@ func (c *Channels) ReceiveDataError(chid datatransfer.ChannelID, err error) erro
...
+// SetDataLimit means a data limit has been set on this channel
+func (c *Channels) SetDataLimit(chid datatransfer.ChannelID, dataLimit uint64) error {
+	c.progressCache.setDataLimit(chid, dataLimit)
+	return c.send(chid, datatransfer.SetDataLimit, dataLimit)
+}
+
+// SetRequiresFinalization sets the state of whether a data transfer can complete
+func (c *Channels) SetRequiresFinalization(chid datatransfer.ChannelID, RequiresFinalization bool) error {
+	return c.send(chid, datatransfer.SetRequiresFinalization, RequiresFinalization)
+}
```
dirkmc, line 385: nit: RequiresFinalization => requiresFinalization

impl/impl.go
```diff
@@ -300,6 +294,50 @@ func (m *manager) SendVoucher(ctx context.Context, channelID datatransfer.Channe
...
+		_ = m.OnRequestDisconnected(channelID, err)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		return err
+	}
+	return m.channels.NewVoucherResult(channelID, voucherResult)
+}
+
+func (m *manager) SetDataLimit(ctx context.Context, chid datatransfer.ChannelID, totalData uint64, stopAtEnd bool) {
+
```
dirkmc, line 338: Looks like the implementation for this function is missing

manager.go
```diff
@@ -8,57 +8,58 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+// ValidationResult describes the result of validating a voucher
+type ValidationResult struct {
+	// Accepted indicates where the request was accepted if a request is not
+	// accepted, the request fails. This is true for revalidation as well
```
dirkmc, line 14:
```suggestion
	// Accepted indicates whether the request was accepted. If a request is not
	// accepted, the request fails. This is true for revalidation as well
```

manager.go
```diff
@@ -8,57 +8,58 @@ import (
...
+	Accepted bool
+	// VoucherResult provides information to the other party about what happened
+	// with the voucher
+	VoucherResult
+	// LeaveRequestPaused indicates whether the request should stay paused
+	// even if the request was accepted
+	LeaveRequestPaused bool
+	// DataLimit specifies how much data this voucher is good for. When the amount
+	// amount data specified is reached (or shortly after), the request will pause
```
dirkmc, line 23:
```suggestion
	// of data specified is reached (or shortly after), the request will pause
```

manager.go
```diff
@@ -8,57 +8,58 @@ import (
...
 		baseCid cid.Cid,
-		selector ipld.Node) (VoucherResult, error)
+		selector ipld.Node) (ValidationResult, error)
 	// ValidatePull validates a pull request received from the peer that will receive data
+	// -- All information about the validation operation is contained in ValidationResult,
+	// including if it was rejected. Information about why a rejection occurred should be embedded
+	// in the VoucherResult.
+	// -- error indicates something went wrong with the actual process of trying to validate
 	ValidatePull(
-		isRestart bool,
```
dirkmc, line 24: IIRC we need to know at the markets layer if the pull request is for a restart. Maybe check when you implement the markets layer changes and revisit.

manager.go
```diff
@@ -8,57 +8,58 @@ import (
...
+type ValidationResult struct {
+	// Accepted indicates where the request was accepted if a request is not
+	// accepted, the request fails. This is true for revalidation as well
+	Accepted bool
+	// VoucherResult provides information to the other party about what happened
+	// with the voucher
+	VoucherResult
+	// LeaveRequestPaused indicates whether the request should stay paused
+	// even if the request was accepted
+	LeaveRequestPaused bool
```
dirkmc, line 21: I suggest renaming this to Paused

statuses.go
```diff
@@ -60,6 +60,18 @@ const (
 	ChannelNotFoundError
 )
 
+func (s Status) IsAccepted() bool {
+	return s != Requested && s != Cancelled && s != Cancelling && s != Failed && s != Failing
```
rvagg, line 64:
```suggestion
	return s != Requested && s != Cancelled && s != Cancelling && s != Failed && s != Failing && s != ChannelNotFoundError
```

?

statuses.go
```diff
@@ -60,6 +60,18 @@ const (
...
+}
+
+func (s Status) IsResponderPaused() bool {
+	return s == ResponderPaused || s == BothPaused || s == Finalizing
+}
+
+func (s Status) InFinalization() bool {
+	return s == Finalizing || s == Completed || s == Completing
```
rvagg, line 72: `ResponderFinalizing` doesn't come into play here at all?

statuses.go
```diff
@@ -60,6 +60,18 @@ const (
 	ChannelNotFoundError
 )
 
+func (s Status) IsAccepted() bool {
+	return s != Requested && s != Cancelled && s != Cancelling && s != Failed && s != Failing
+}
+
+func (s Status) IsResponderPaused() bool {
+	return s == ResponderPaused || s == BothPaused || s == Finalizing
```
rvagg, line 68: `ResponderFinalizing`?

channels/channels_test.go
```diff
@@ -168,61 +168,108 @@ func TestChannels(t *testing.T) {
...
+		_ = checkEvent(ctx, t, received, datatransfer.DataQueuedProgress)
+		state = checkEvent(ctx, t, received, datatransfer.DataQueued)
+		require.Equal(t, uint64(300), state.Queued())
+
+		err = channelList.SetDataLimit(datatransfer.ChannelID{Initiator: peers[1], Responder: peers[0], ID: tid1}, 400)
+		require.NoError(t, err)
+		state = checkEvent(ctx, t, received, datatransfer.SetDataLimit)
+		require.Equal(t, state.DataLimit(), uint64(400))
+
+		// send block again has no effect
```
rvagg, line 241: so .. can you correct me if my interpretation is wrong about what the sequence of initial steps here look like - setting a data limit applies to the entirety of the channel, including data that's already been received, so the `400` limit includes the `300` that's already been received. Then, in this next step, the same data is re-queued but not counted because it's already been counted (aside: why are we able to receive the same data again, and why don't we count duplicate data against the limit? is there abuse possibilities hiding in here?). It's when we get to the next `200` that we go over the `400` and hit a pause. Then the `700` limit set further down applies again to the whole channel, from the start, including the `400` already queued.

Sound about right?

channels/channels.go
```diff
@@ -362,29 +392,74 @@ func (c *Channels) HasChannel(chid datatransfer.ChannelID) (bool, error) {
...
+		return false, false, err
+	}
+
+	// if no data progress, return
+	if !progress {
+		return
+	}
+
+	// if we don't check data limits on this function, return
+	if readProgress == nil {
```
rvagg, line 456: what are the cases where this is `nil`? is it just `DataSent` calls? and if so, the only value in calling this function is to set the progress of the sent index, correct?

very minor nit - the naming of this function could be improved, current name doesn't imply mutation but that's one of its two core functions, `updateIndexAndCheckEvents()` might be the verbose option.

channels/channels.go
```diff
@@ -362,29 +392,74 @@ func (c *Channels) HasChannel(chid datatransfer.ChannelID) (bool, error) {
...
+		// return a pause error so the transfer knows to pause
+		return datatransfer.ErrPause
+	}
+	return nil
+}
+
+func (c *Channels) checkEvents(chid datatransfer.ChannelID, evt datatransfer.EventCode, delta uint64, index int64, unique bool, readFromOriginal readIndexFn, readProgress readProgressFn) (pause bool, progress bool, err error) {
+
+	// if this is not a unique block, no data progress is made, return
+	if !unique {
```
rvagg, line 440: and .. if `!unique` we're certain that the index doesn't need updating? because that's a new shortcut over the original

impl/events.go
```diff
@@ -220,59 +162,78 @@ func (m *manager) OnResponseReceived(chid datatransfer.ChannelID, response datat
...
-		if response.IsNew() {
-			log.Infof("channel %s: received new response, accepting channel", chid)
-			err := m.channels.Accept(chid)
-			if err != nil {
-				return err
-			}
+	}
+
+	// was this the first response to our initial request
+	if response.IsNew() {
```
rvagg, line 175: this is interesting, both IsNew and IsRestart jumping outside of the previously containing `if`, was it wrong before not to check this in all but the IsComplete case?

channels/caches.go
```diff
@@ -0,0 +1,133 @@
...
+type readIndexFn func(datatransfer.ChannelID) (int64, error)
+
+type cacheKey struct {
+	evt  datatransfer.EventCode
+	chid datatransfer.ChannelID
+}
+
+type blockIndexCache struct {
+	lk     sync.RWMutex
+	values map[cacheKey]*int64
```
rvagg, line 19: out of interest: how do we manage blocks as _indexes_, is that something we defer to the protocol and we just record a count? if we can have duplicate indexes, is that something we care about here or does that somehow bubble up from the protocol?

manager.go
hannahhoward: ultimately changed to ForcePause in subsequent PR

statuses.go
hannahhoward: ResponderFinalizing is the for the initiator, indicating the responder is finalizing. For the moment, we only care about the responder being in a finalization state, as it only affects their logic.

Probably and maybe we should add it to the list but I just want to be careful.

statuses.go
hannahhoward: that's fair.

statuses.go
hannahhoward: I think I'll just go ahead and add it

channels/channels_test.go
hannahhoward: yes

channels/channels.go
hannahhoward: yea agree on all fronts. I'm inclined to wait and update when the transport refactor happens, cause I think data limit logic will move downward to the transport

channels/channels.go
hannahhoward: unique is again something that really should go away at the data transfer level -- the transport should only call these when unique == true.... but that's for the next refactor

impl/events.go
hannahhoward: honestly they're just unneccesarily nested. IsNew = true & IsRestart = tree imply IsValidationResult == true, as does IsComplete = true.

So basically, the IsValidationResult handles the "!Accepted" condition and then the other things are determining behavior on an accepted response.

channels/caches.go
hannahhoward: we record a count from the protocol. honestly I'm a bit ambiguous about indexes in general filtering up to data transfer but I think for now I prefer it.

PR #322, opened 2022-05-03
Refactor revalidators v2

hannahhoward:
# Goals

This PR was written during the course of implementing the voucher refactor in go-fil-markets. 

# Implementation

The main goal in refactoring the go-fil-markets retrieval provider was to drive towards clear, easy to reason about logic with very clear event structure and idempotent code that restarts easily.

What became clear is that fundamentally, voucher revalidation needs to happen asynchronously. So now, the consuming library simply monitors data transfer events and updates validation status as new vouchers come in. We changed Revalidate to ValidateRestart -- so that it only covers restarts, while removing synchronous revalidation entirely. Now, we simply save new vouchers, and we've added an UpdateValidationStatus method to change validation parameters when the markets software processes all vouchers.

Moreover, the idea of Pausing in validation needs to be separated from DataLimits and Finalization. The way to unpause a request held on a data limit is to increase the data limit. The way to end finalization is to set RequiresFinalization to false. The pause property is renamed from LeaveRequestPaused to ForcePause, clearly identifying that the markets library is explicitly applying it's own logic around pausing -- in this case, pausing to Unseal.

Finally I did a minor reorderng of validation events to promote more predictable behavior near the end of a transfer (not accidentally completing the channel before events get processed)

### DIFF
--- a/impl/impl.go
+++ b/impl/impl.go
@@ -368,7 +368,7 @@ func (m *manager) updateValidationStatus(ctx context.Context, chid datatransfer.
 func (m *manager) processValidationUpdate(ctx context.Context, chid datatransfer.ChannelID, result datatransfer.ValidationResult) (datatransfer.ChannelState, datatransfer.Response, error) {
 
 	// read the channel state
-	chst, err := m.channels.GetByID(context.TODO(), chid)
+	chst, err := m.channels.GetByID(ctx, chid)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/impl/impl.go
+++ b/impl/impl.go
@@ -24,6 +24,7 @@ import (
 	"github.com/filecoin-project/go-data-transfer/v2/channels"
 	"github.com/filecoin-project/go-data-transfer/v2/encoding"
 	"github.com/filecoin-project/go-data-transfer/v2/message"
+	"github.com/filecoin-project/go-data-transfer/v2/message/types"
 	"github.com/filecoin-project/go-data-transfer/v2/network"
 	"github.com/filecoin-project/go-data-transfer/v2/registry"
 	"github.com/filecoin-project/go-data-transfer/v2/tracing"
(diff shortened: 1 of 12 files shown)


impl/impl.go
```diff
@@ -334,8 +335,108 @@ func (m *manager) SendVoucherResult(ctx context.Context, channelID datatransfer.
...
+	chst, response, err := m.processValidationUpdate(ctx, chid, result)
+
+	// dispatch transport updates
+	return m.handleTransportUpdate(ctx, chst, response, result, err)
+}
+
+func (m *manager) processValidationUpdate(ctx context.Context, chid datatransfer.ChannelID, result datatransfer.ValidationResult) (datatransfer.ChannelState, datatransfer.Response, error) {
+
+	// read the channel state
+	chst, err := m.channels.GetByID(context.TODO(), chid)
```
rvagg, line 371:
```suggestion
	chst, err := m.channels.GetByID(ctx, chid)
```

impl/impl.go
```diff
@@ -334,8 +335,108 @@ func (m *manager) SendVoucherResult(ctx context.Context, channelID datatransfer.
...
+	if resultErr == nil && result.Accepted && !pauseRequest {
+		if chst.Status().IsResponderPaused() && !chst.Status().InFinalization() {
+			return m.transport.(datatransfer.PauseableTransport).ResumeChannel(ctx, response, chst.ChannelID())
+		}
+	}
+
+	// send a response message
+	if response != nil {
+		if err := m.dataTransferNetwork.SendMessage(ctx, chst.ChannelID().Initiator, response); err != nil {
+			return err
```
rvagg, line 424: is it appropriate that we continue executing this function if we successfully send the message? can we have a `SendMessage` as well as a `CloseChannel` or `PauseChannel`?

rvagg (approved):
looks OK from a superficial perspective and the rationale makes sense

it's just a dramatic change; I'll have to take these two big refactors into account in my work which has already diverged quite a bit from `master`

impl/impl.go
hannahhoward: yes, we can, and the reason resume gets singled out is cause of the current state of assumptions about the transport. honestly, the transport refactor should make this look a lot more legible.

PR #324, opened 2022-05-19
Refactor transport part 1

hannahhoward:
# Goals

Well, I got to passing integration tests with the messaging layer underneath the transport. I'm not feeling done at all yet, truthfully. The question now is whether to try to get all tests passing and merge the changeset, or push for something closer to what I think makes sense before we go forward.

The biggest issue at the moment is that the "dtChannel" struct, which grew sort of like a frankenstein to begin with, is even more of a frankenstein now. I think I need to dig into refactoring this. The whole of the transport code is very messy. While I can certainly finish off the unit tests, I don't know what value we get from unit testing some very rough code.


hannahhoward:
Proposed path forward: get unit tests outside the transport passing, comment out transport tests for now, fix transport code in a subsequent PR.

hannahhoward:
So: I'm asking @rvagg and maybe @dirkmc or @willscott to look at the code outside the transport layer with an unfortunate promise for the moment that the code "just works" by way of the integration tests, understanding the merge is to V2, not to master, and there will be subsequence prs before release. (though to be fair there will be additional changes made to the transport boundary for the events handlers.

hannahhoward:
Things I am most interested in review on:
1. The current transport interface
2. The "WithTransport" method
3. Code changes in the `impl` directory (as opposed to the transport dir)

channels/channel_state.go
```diff
@@ -110,32 +116,54 @@ func (c channelState) Message() string {
...
+	for i, encoded := range c.ic.Vouchers {
+		if c.decodedVouchers[i] != nil {
+			vouchers = append(vouchers, *c.decodedVouchers[i])
+			continue
+		}
 		decoder, _ := c.voucherDecoder(encoded.Type)
 		encodable, _ := decoder.DecodeFromCbor(encoded.Voucher.Raw)
-		vouchers = append(vouchers, encodable.(datatransfer.Voucher))
+		voucher := encodable.(datatransfer.Voucher)
+		c.decodedVouchers[0] = &voucher
```
rvagg, line 127:
```suggestion
		c.decodedVouchers[i] = &voucher
```

looks like a copypasta bug?

impl/restart.go
```diff
@@ -123,12 +58,12 @@ func (m *manager) openPullRestartChannel(ctx context.Context, channel datatransf
...
 		transportConfigurer(chid, voucher, m.transport)
 	}
-	m.dataTransferNetwork.Protect(requestTo, chid.String())
 
 	// Monitor the state of the connection for the channel
-	monitoredChan := m.channelMonitor.AddPullChannel(chid)
-	log.Infof("sending open channel to %s to restart channel %s", requestTo, chid)
-	if err := m.transport.OpenChannel(ctx, requestTo, chid, cidlink.Link{Cid: baseCid}, selector, channel, req); err != nil {
+	monitoredChan := m.channelMonitor.AddChannel(chid, channel.IsPull())
+	log.Infof("sending push restart channel to %s for channel %s", requestTo, chid)
```
rvagg, line 64:
```suggestion
	log.Infof("sending restart channel to %s for channel %s", requestTo, chid)
```

rvagg:
OK, so we're going with the slightly odd `WithChannel()` actions callback API because different transports may have different "capabilities", so the actions object you get may have different methods on it depending on the transport. I suppose that's reasonable, but some questions to further clarify whether this is really needed:

1. If you can query `transport.Capabilities()` prior to even touching `WithChannel()`, couldn't we just have a single actions object that can be constructed without the need for the actions callback, then pass that in and have it error if you requested actions that the transport doesn't have the capabilities for? i.e. is the actions callback really needed and couldn't it just be done with a declarative config object passed in to `WithChannel()`.
2. What variations in capabilities are you designing for here? I imagine this is mostly about thinking ahead to the features that bitswap doesn't support that Graphsync does, but maybe you're also thinking broader to other protocols?

hannahhoward:
@rvagg OK I've modified my interfaces significantly, based on your feedback. I still have a kind of "multi-command" in the form of UpdateChannel, but all the individual methods are also here.

Would love your feedback. Also, the transport code should be generally easier to comprehend now.

impl/impl.go
```diff
@@ -503,24 +443,21 @@ func (m *manager) CloseDataTransferChannelWithError(ctx context.Context, chid da
...
-	err = m.transport.CloseChannel(ctx, chid)
-	if err != nil {
-		log.Warnf("unable to close channel %s: %s", chid, err)
-	}
-
-	// Try to send a cancel message to the remote peer. It's quite likely
-	// we aren't able to send the message to the peer because the channel
-	// is already in an error state, which is probably because of connection
-	// issues, so if we cant send the message just log a warning.
+	// Close transfport and try to send a cancel message to the remote peer.
```
rvagg, line 446:
```suggestion
	// Close transport and try to send a cancel message to the remote peer.
```

testutil/faketransport.go
```diff
@@ -74,22 +129,11 @@ func (ft *FakeTransport) SetEventHandler(events datatransfer.EventsHandler) erro
 	return ft.SetEventHandlerErr
 }
 
+// Shutdownc loses a this transport
```
rvagg, line 132:
```suggestion
// Shutdown loses a this transport
```

rvagg:
Nice. I haven't quite gone thoroughly through the graphsync bits yet, lots of detail there. But the API is nicer and much more direct.

Two questions:

1. Could we get rid of `PauseChannel()`, `CloseChannel()` and `SendMessage()` entirely and end up with a simplified API, where you need to go through `UpdateChannel()` for all of those operations? It might also encourage usage that batches operations if you don't offer the atomic forms.
2. Is there a case for making `ChannelUpdate#SendUpdate` a slice instead of a single `Message` so you could send multiple messages? Or are we not dealing with communication channels that have a case for multiple messages without response?

impl/impl.go
```diff
@@ -399,7 +399,10 @@ func (m *manager) CloseDataTransferChannel(ctx context.Context, chid datatransfe
 	))
 	defer span.End()
 	// Close the channel on the local transport
-	err = m.transport.CloseChannel(ctx, chid)
+	err = m.transport.UpdateChannel(ctx, chid, datatransfer.ChannelUpdate{
+		Paused: chst.Status().IsResponderPaused(),
```
rvagg, line 403: yeah, this is slightly awkward isn't it? as long as we have access to the channel status this then it's probably not the worst thing

you could make it an enum with 3 values: pause, unpause (resume?), continue - where the latter is "just keep on doing what you were doing".

PR #325, opened 2022-05-20
feat(ipld): vouchers as plain ipld.Node

rvagg:
Yet another take at this; previous at #305 which is impossible to rebase now and is probably not correct anyway.

This isn't done, tests aren't passing yet; still some problems in here that need to be figured out but I'll tackle it again next week.


rvagg:
I managed to fix some of the integration test failures, but there's a few remaining I haven't figured out yet.

The problem I was having is in needing to have two things for a voucher now instead of one - you need the voucher (`ipld.Node`) and the voucher type (`datamodel.TypeIdentifier`). So there's some places where formerly the voucher is just passed around but I've neglected to pass through the type as well. Fixing the ones I could find helped with most of the failures. I don't know if the rest are related.

rvagg:
Hah! I figured it out, got to green. And it was related to voucher types.

The API is now more verbose because of the voucher:voucherType pairs everywhere. The main places where this becomes painful from the outside is the additional arguments to `OpenPullDataChannel()` and `OpenPushDataChannel()`. In those two calls you need to provide a voucher, voucherType _and_ voucherResultType. The two types are cached on the internal channel and are used for future interactions. Then there's another change in `acceptRequest()` that also sets up a new channel. It gets the voucher and voucherType from the incoming message and the voucherResultType from the `validatorFunc()` call. Like everywhere else, `ValidationResult`s need to provide the pair, this time a VoucherResult and a VoucherResultType.

An alternative might be to make a new type to deal with this verbosity and use it everywhere:

```go
type Voucher {
  Data ipld.Node
  Type datatransfer.TypeIdentifier
}
```

Aside from the verbosity, I don't really know if this API makes sense in all places it gets used:

1. Do we have this information available at all of these places? Is the voucher type always a known quantity?
2. Is it true that both `voucherType` and `voucherResultType` are stable across the lifetime of a channel? Does a channel only have one type of each, or can they vary? The code being changed relied on getting the `Type()` off the vouchers, but it also seemed to be able to make an assumption that the first voucher's `Type()` is reflective of all of the vouchers in that channel.

hannahhoward:
My initial gut reaction is we should keep the type near the voucher, so I like the proposed type. As I read the current code, there is a bit of loss of type information as the current voucher interface always has the Type() method, and the Type is encoded into the serialized data. My biggest immediate concern is we are changing the serialized data for transfers in a way we need we need to write migrations for -- it's find to press forward, but the symmetry between the current EncodedVoucher and your proposed Voucher struct suggests we're on the right track. 

Vouchers are not consistent across a whole transfer. For example, retrievals use both DealProposal and DealPayment as voucher types.

message/message1_1prime/message.go
```diff
@@ -219,16 +181,16 @@ func FromNet(r io.Reader) (datatransfer.Message, error) {
 }
 
 // FromNet can read a network stream to deserialize a GraphSyncMessage
-func FromIPLD(node datamodel.Node) (datatransfer.Message, error) {
+func FromIPLD(node ipld.Node) (datatransfer.Message, error) {
```
hannahhoward, line 184: My understanding is that datamodel.Node was the "updated" place to refer to these as the rest were just type aliases for compatibility with older versions.

ipldutils/ipldutils.go
```diff
@@ -0,0 +1,169 @@
+package shared
+
```
hannahhoward, line 2: What is going on with all this code at this point? It seems like a bunch of complexity maybe designed to support actually picking bindnode schemas when deserializing data transfer vouchers, but it's only used for the actual data transfer message formats. Do we need it? does this actually make the message code easier to read?

impl/impl.go
```diff
@@ -279,7 +273,7 @@ func (m *manager) SendVoucher(ctx context.Context, channelID datatransfer.Channe
 		span.SetStatus(codes.Error, err.Error())
 		return err
 	}
-	updateRequest, err := message.VoucherRequest(channelID.ID, voucher.Type(), voucher)
+	updateRequest, err := message.VoucherRequest(channelID.ID, chst.VoucherType(), voucher)
```
hannahhoward, line 276: This is an example that will fall down in the context of actual retrieval -- these types may not be the same. On the other hand, based on our modified markets code, it probably does not matter that much.

hannahhoward (commented):
Some initial comments. I definitely think the struct w/ a type is a good idea.

ipldutils/ipldutils.go
rvagg: TODO - it's copypasta from go-fil-markets code currently which is evolving to deal with message complexity there. But it's probably way too much here.

My thought was to extract this to a separate package but let's discuss this. The idea was that we could hide the schemas within the type itself so you just need to pass around type pointers and you don't need to pre-register for prototypes, it's done lazily and cached...

Mostly I'd really like to be using similar patterns across both codebases for this bindnode work, somehow, if possible.

message/message1_1prime/message.go
rvagg: yeah, I don't think it matters too much but Eric would probably be happier if I didn't use it like this.

manager.go
```diff
@@ -101,29 +113,29 @@ type Manager interface {
 	// RegisterVoucherType registers a validator for the given voucher type
 	// will error if voucher type does not implement voucher
 	// or if there is a voucher type registered with an identical identifier
-	RegisterVoucherType(voucherType Voucher, validator RequestValidator) error
+	RegisterVoucherType(voucherType TypeIdentifier, validator RequestValidator) error
 
 	// RegisterVoucherResultType allows deserialization of a voucher result,
 	// so that a listener can read the metadata
-	RegisterVoucherResultType(resultType VoucherResult) error
+	RegisterVoucherResultType(resultType TypeIdentifier) error
```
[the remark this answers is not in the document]
rvagg: :thumbsup: yeah I was wondering if it was OK to remove this, have done it now

rvagg:
OK, this is getting closer I think. `datatransfer.TypedVoucher` is a tuple that binds a `Node` and a `TypeIdentifier` together and is now required through most of the API. There's some internal places where a `Node` will be by itself - they're decoupled inside messages (we could couple them there too, but it's not so important right now) and in validators - because you register a validator for a type so you probably don't need to see that type again, right?

Channels should now be able to support multiple voucher types (although this still isn't tested in here, but will be by go-fil-markets); and the serialization of the statemachine should be the same as before, I've reverted all changes to the channels/internal package.

impl/receiving_requests.go
```diff
@@ -251,8 +265,8 @@ func (m *manager) recordAcceptedValidationEvents(chst datatransfer.ChannelState,
 	chid := chst.ChannelID()
 
 	// record the voucher result if present
-	if result.VoucherResult != nil {
-		err := m.channels.NewVoucherResult(chid, result.VoucherResult)
+	if result.VoucherResult != nil && result.VoucherResult.Voucher != nil {
```
hannahhoward, line 268: Why would someone pass a VoucherResult that wasn't nil but had a nil voucher? I guess that's a possibility and perhaps best to be defense against code supplied outside the library.

types.go
```diff
@@ -74,10 +68,10 @@ type Channel interface {
 
 	// Selector returns the IPLD selector for this data transfer (represented as
 	// an IPLD node)
-	Selector() ipld.Node
+	Selector() datamodel.Node
 
-	// Voucher returns the voucher for this data transfer
-	Voucher() Voucher
+	// Voucher returns the initial voucher for this data transfer
+	Voucher() (TypedVoucher, error)
```
hannahhoward, line 74: I think this is ultimately the correct change and something we should have done a while ago. That said, keep in mind this will bubble up a lot of api changes. We had better factor in integration with Lotus + Filclient + Boost at a minimum in our timelines

hannahhoward (approved):
This is honestly looking really good. I'd like to go ahead and merge to v2 and you can cleanup in subsequent PRs.

PR #326, opened 2022-05-21
Do not EncodeToNode selectors

hannahhoward:
# Goals

Cut down memory usage -- see here from a recent test in AutoRetrieve:
profile001

Looks like our EncodeToNode function is eating memory (somewhat predictably since we knew it was a hack). However, on the graph 1.35GB is going through https://github.com/filecoin-project/go-data-transfer/blob/master/message/message1_1prime/message.go#L30, which is an EncodeToNode call on a selector. Given a selector is already an ipld.Node, we don't need this.

# Implementation

Remove round trip for selector.

### DIFF
--- a/message/message1_1prime/message.go
+++ b/message/message1_1prime/message.go
@@ -27,11 +27,6 @@ func NewRequest(id datatransfer.TransferID, isRestart bool, isPull bool, vtype d
 		return nil, xerrors.Errorf("base CID must be defined")
 	}
 
-	selnode, err := encoding.EncodeToNode(selector)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


rvagg (approved):
So, the only thing this doesn't do is the checked for `TypedNode` and conversion to `Representation()`; and, I'm not all that sure what's going on underneath if we get a `TypedNode` selector at this point via bindnode.

It's very unlikely we'll get a `TypedNode` here; I don't think we even have a proper `TypedPrototype` for selectors, it's all manual still, but maybe we can't rule it out?

PR #335, opened 2022-06-15
Fix transfer message bindnode wrap passing pointer to pointer

masih:
Bindnode wrap expects a pointer. A fail-fast check is introduced in
ipld/go-ipld-prime#427 which will cause a panic if the IPLD dependency
is upgraded to the head of main branch.

Avoid passing pointer-to-pointer when wrapping transfer message.

See:
- https://github.com/ipld/go-ipld-prime/pull/427

### DIFF
--- a/message/message1_1prime/transfer_message.go
+++ b/message/message1_1prime/transfer_message.go
@@ -30,7 +30,7 @@ func (tm *TransferMessage1_1) TransferID() datatransfer.TransferID {
 }
 
 func (tm *TransferMessage1_1) toIPLD() schema.TypedNode {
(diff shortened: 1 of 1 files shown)


hannahhoward (approved):
I'm pretty sure this has changed significantly (and been fixed?) in v2 (our current workstream) but we should merge this for compatibility on v1 with latest go-ipld-prime.

hannahhoward:
cc: @rvagg for final approval

rvagg (approved):
yeah, fixed and revamped over in the new branch @ https://github.com/filecoin-project/go-data-transfer/pull/323, but if this is being used on master with the new go-ipld-prime then I guess this is fine to merge

PR #336, opened 2022-06-15
Use CborGenCompatibleNode

hannahhoward:
# Goals

While working transport changes I discovered why I think is a pretty easy improvement to how we serialize channel state, by copying @rvagg 's CborGenCompatibleNode from https://github.com/filecoin-project/go-fil-markets/pull/713

# Implementation

- replace cbg.Deferred's from internal channel state with CborGenCompatibleNode
- this allows to immediately deserialize to datamodel.Node with basicnode.
- this means we can drop all the error returns from the channelState interface
- I think this makes sense, as it moves errors to the deserialization step, IMHO where they belong.


rvagg (approved):
yeah, nice, removing those errors is good

PR #342, opened 2022-06-30
Refactor events handlers

hannahhoward:
# Goals

Complete the transport refactor, moving several items below the transport boundary, cleaning up event flow, fixing issues in the state machine, making transport interface more high level, and limit data transfer manager's responbilities to voucher processing and state handling.

# Implementation

1. Cleaning up the initiating states flow

- We've added a "TransferInitaited" step for when the request begins processing. We've removed the "TransferRequestQueued". TransferInitaited is called when a graphsync (or other transport) actually begins executing a request.
- Recently, we merged https://github.com/ipfs/go-graphsync/pull/384. We now evaluate whether to accept requests immediately, then queue them for later processing. Previously we had queued for processing without evaluating whether to ultimately accept a request.
- Combined with the new event, this means a high bandwidth data sender using graphsync will see a Requested -> Queued -> Ongoing sequence, where Queued indicates a request has been accepted but we are waiting for it to reach the top of the graphsync processing queue.  This is a long requested feature to be able to know when requests have been received and accepted, even when they make take a long time to start processing. Additionally rejected requests will fail much faster
- What if we receive TransferInitaited before Accept? In a graphsync pull request, the graphsync requestor will begin "processing" their request -- producing a TransferIntiaited -- and then only later receive a reply back from the server accepting the request. For this sequence (first TransferInitated then Accept), the intermediate state is called AwaitingAcceptance.

2. Cleaning up pause / resumes and the state machine.

- Previous the state machine's process for coming out of Pause/Resume was simply to go to Ongoing, regardless of the state it had been in before pause resume. The proper behavior would be to return to the state it was in before the pause / resume. While we've gotten away with it largely, there are some rough behaviors that emerge from it.
- Essentially, pausing and resuming is not great for an FSM where state is represented by a single value.  In FSM terminology Pause/Resume is what I believe is referred to as a "orthogonal region" (though possibly a hierarchical state machine... who knows). 
- I briefly explored adding more features to go-statemachine then took the simple path of simply not making pause resume part of the state name, and rather seperate attributes on the channel state.

3. Moving data limit tracking into the transport

There was a ton of awful logic in the channel state machine plus some memory caches trying to track which new "I sent/received data" events from the transport were actually new, as well as when exactly when we hit our data limits so we can tell the transport to pause. The really simple solution here is the transport should just handle this. It now does. In fact, we could make this even simpler and remove the "DataSentProgress" / "DataQueuedProgress" / "DataReceivedProgress" events because we now know for sure that they will always happen one-to-one with the corresponding "DataSent/DataQueued/DataReceived". I've left them there for now cause existing code may be expecting these. But, also, maybe just fuck-it -- let's make it simple! 

Similarly, the transport now tracks data limits and pauses itself as needed.

4. Making index tracking abstract

Previously, every piece of data sent (i.e. each block) had an integer index. I did a lot of thinking about what an "index" actually is, and I finally came to the conclusion it's simply the transports way of knowing how to pick up again after a restart. In the case of graphsync and selectors, I think the ultimate ideal way to restart is "restart form this path in a selector traversal" (hey this even exists actually already -- https://github.com/ipld/go-ipld-prime/pull/358). I thought about what I would do to eventually evolve the graphsync transport to eventually do resume at path, and I realized I'd need to switch the index type. So with that in mind, I decided to just make indexes IPLD nodes. The value of the IPLD here is simply that it can be serialized into the transfer state. Otherwise it's effectively type "any", with the transport being the one that knows how to interpret it.

5. The new transport events interface

I went through several iterations on the transport events handler interface before finally settling on what I ended up with. 

I wanted to distinguish between "blocking" events -- where the transport expects a response from go-data-transfer -- and non-blocking events that could potentially be dispatched asynchronously, and to limit the number of blocking events. I attempted to make almost everything asychronous at one point, then pulled back a bit at the end. I ultimately decided new requests and responses needed to be processed and return errors as needed, because the transport uses those errors to accept or reject requests. Everything else is now dispatched through an OnTransportEvent() callback, which has no return value. While it's actually called synchronously for the moment, theoretically it could be asynchronous, though attention would need to be paid to ordering with the OnRequestReceived/OnResponseReceived synchronous events.

I also went through several iterations as well trying to think about how data transfer would pass on things like adjusted data limits and updated pause / resume states. I went through various interations of a "channel update" data structure building on the one I had in earlier transport refactor PRs. I ultimately decided the absolute simpest thing was to let go-data-transfer tell the transport when state had been updated and let the transport simply pull ChannelState as needed. The resulting code so far seems relatively reasonable.

6. I fixed the flaky integraiton test with pause/resume

This was driving me crazy

7. I removed errors on a bunch of the message creation functions, cause they never returned erros

It's a simple way to clean up code

8. I redid the testing infrastructure for the transport tests and wrote new ones

Transport unit testing is really weird without a real graphsync below or a responsive data transfer above. But the current versions seem slightly more comprehesible?

# For Discussion

- TODO before merge:
   - the transport tests need error cases added. I left the old tests commented out in one of the files for reference
   - there are some legacy "Transport Events" from experiments with making more stuff asynchronous commented out
   - I'm not sure the state store migrations I wrote cover everything -- I need to review them cause I wrote them early but I think I made more changes since then
   
- TODOS before data transfer v2.0 shipping
   - This PR deprecates some events, and there are more already deprecated. Seems like a good time to actually delete them, given v2 is intended to be a breaking code change (but not protocol breaking). Unfortunately, currently they are a numbered sequence, which suggests a more radical and much needed change -- making our events and states string based. This would save so many people so much trouble. I have no idea why they're number based other than when they were being developed before launch people actually thought these numbers would have great significance.
 
 - Open Questions
    - Was it worth it?


hannahhoward:
Oh one more comment -- the code in the graphsync transport is still messy, and there are various things I'd like to consider changing there. However, the good news is, if the interfaces are right, we clean this up later as a non-breaking change.

hannahhoward:
One more proposal, which could be a challenge, but which I think is ideal in the long term, is have everything be asychronous -- send commands down, events up. It's a bit tricky to coordinate with graphsync, but boy this would make everything clearer in the logic.

rvagg:
> I fixed the flaky integraiton test with pause/resume

yay

channels/internal/migrations/migrations.go
```diff
@@ -1,13 +1,111 @@
...
+		Status:               oldChannelState.Status,
+		Queued:               oldChannelState.Queued,
+		Sent:                 oldChannelState.Sent,
+		Received:             oldChannelState.Received,
+		Message:              oldChannelState.Message,
+		Vouchers:             oldChannelState.Vouchers,
+		VoucherResults:       oldChannelState.VoucherResults,
+		ReceivedIndex:        internal.CborGenCompatibleNode{Node: receivedIndex},
+		SentIndex:            internal.CborGenCompatibleNode{Node: sentIndex},
+		QueuedIndex:          internal.CborGenCompatibleNode{Node: queuedIndex},
```
rvagg, line 98: I imagine the thinking about adding migration testing is that this transition is pretty straightforward and the absolute hassle of having to come up with fixtures for this is just not worth it? It seems good from eyeballing it but we'll probably want to see it actually working in practice.

channels/internal/migrations/migrations_cbor_gen.go
```diff
@@ -0,0 +1,876 @@
+// Code generated by github.com/whyrusleeping/cbor-gen. DO NOT EDIT.
```
rvagg, line 1: Does the diff of this with the previous internalchannel_cbor_gen.go show no meaningful change?

channels/channels_fsm.go
```diff
@@ -160,57 +164,65 @@ var ChannelEvents = fsm.Events{
...
+	// seems less than ideal. We need some kind of support for pausing being an independent aspect of state
+	// Possibly we should just remove whether a state is paused from the state entirely.
 	fsm.Event(datatransfer.PauseInitiator).
-		FromMany(datatransfer.Requested, datatransfer.Ongoing).To(datatransfer.InitiatorPaused).
-		From(datatransfer.ResponderPaused).To(datatransfer.BothPaused).
-		FromAny().ToJustRecord().Action(func(chst *internal.ChannelState) error {
-		chst.AddLog("")
-		return nil
-	}),
+		FromMany(datatransfer.Ongoing, datatransfer.Requested, datatransfer.Queued, datatransfer.AwaitingAcceptance).ToJustRecord().
```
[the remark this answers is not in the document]
hannahhoward: yea exactly. I'm not convinced FSM is the right model in a larger sense anyway, so I'd rather not further invest in even more complicated options.

channels/internal/migrations/migrations.go
hannahhoward: no I should probably do that!

channels/internal/migrations/migrations_cbor_gen.go
hannahhoward: probably? it's all auto generated scripts but the structs are copy paste.

transport.go
```diff
@@ -17,75 +17,118 @@ const LegacyTransportID TransportID = "graphsync"
...
+}
+
+type TransportReceivedCancelRequest struct {
+	Request Request
+}
+
+// TransportReceivedResponse occurs when we receive a response to a request
+type TransportReceivedResponse struct {
+	Response Response
+}
```
rvagg, line 71: what's the story with these? are they are TODO in order to finish this work? they seem like important pieces of the puzzle

rvagg: ah, I think these come under:

> there are some legacy "Transport Events" from experiments with making more stuff asynchronous commented out

But I'd like to hear more about these specific events - were they never actually used? Are we losing anything by not having them?

impl/impl.go
```diff
@@ -331,134 +321,122 @@ func (m *manager) updateValidationStatus(ctx context.Context, chid datatransfer.
...
 	}
 
-	// if the request is now rejected, error the channel
-	if !result.Accepted {
-		err = m.recordRejectedValidationEvents(chid, result)
-	} else {
-		err = m.recordAcceptedValidationEvents(chst, result)
-	}
+	// dispatch channel events and generate a response message
+	err = m.processValidationUpdate(ctx, chst, result)
```
rvagg, line 331: Not sure this refactor buys you anything tbh, it's fine, but how much I miss ternaries in Go! In a sane programming language this would simply be:

```go
err = !result.Accepted ? m.recordRejectedValidationEvents(chid, result) : m.recordAcceptedValidationEvents(chst, result)
```

itest/integration_test.go
```diff
@@ -1494,8 +1495,27 @@ func TestPauseAndResume(t *testing.T) {
 			sv := testutil.NewStubbedValidator()
 			sv.StubResult(datatransfer.ValidationResult{Accepted: true})
 			sv.StubRestartResult(datatransfer.ValidationResult{Accepted: true})
-
 			var chid datatransfer.ChannelID
+
+			gsData.Gs1.RegisterOutgoingBlockHook(func(p peer.ID, r graphsync.RequestData, block graphsync.BlockData, ha graphsync.OutgoingBlockHookActions) {
+				if block.Index() == 5 && block.BlockSizeOnWire() > 0 {
```
rvagg, line 1501: so the flake in here was a race on where the pause was triggered? and shifting further down the pipe to the hook, with added delay, fixed it?

message/message1_1prime/message.go
```diff
@@ -76,7 +76,7 @@ func UpdateRequest(id datatransfer.TransferID, isPaused bool) datatransfer.Reque
 }
 
 // VoucherRequest generates a new request for the data transfer protocol
-func VoucherRequest(id datatransfer.TransferID, voucher *datatransfer.TypedVoucher) (datatransfer.Request, error) {
```
rvagg, line 79: 👌

statuses.go
```diff
@@ -58,18 +60,79 @@ const (
 
 	// ChannelNotFoundError means the searched for data transfer does not exist
 	ChannelNotFoundError
+
+	// Queued indicates a data transfer request has been accepted, but is not actively transfering yet
+	Queued
+
+	// AwaitingAcceptance indicates a transfer request is actively being processed by the transport
+	// even if the remote has not yet responded that it's accepted the transfer
```
rvagg, line 68:
```suggestion
	// even if the remote has not yet responded that it's accepted the transfer. Such a state can
	// occur, for example, in a requestor-initiated transfer that starts processing prior to receiving
	// acceptance from the server.
```

or something like that, or even get specific and talk about graphsync pull reuqests; because this one's a bit of a mystery without the kind of clarification you gave in your OP.

transport/graphsync/dtchannel/dtchannel.go
```diff
@@ -262,6 +282,113 @@ func (c *Channel) Resume(ctx context.Context, extensions []graphsync.ExtensionDa
...
+func (c *Channel) UpdateFromChannelState(chst datatransfer.ChannelState) error {
+	c.lk.Lock()
+	defer c.lk.Unlock()
+	return c.updateFromChannelState(chst)
+}
+
+func (c *Channel) updateFromChannelState(chst datatransfer.ChannelState) error {
+	// read the sent value
+	sentNode := chst.SentIndex()
+	if !sentNode.IsNull() {
```
rvagg, line 305: More of a comment than a change request: these `IsNull`s dictate the contract on `ChannelState` that is currently fulfilled by `CborGenCompatibleNode`. If we were to move away from cbor-gen then that contract would need to follow. By default the `Node`s present here would need to satisfy `IsNull()==true` in their default state—which isn't necessarily natural (e.g. maybe `Absent` or simply `nil` would be more natural choices).

:shrug: I think I'm just feeling the mild awkwardness of the implicit `datatransfer.ChannelState` contract being dictated by use this far down in the stack. Hopefully tests would pick up any problems if we changed later on without realising this.

transport/graphsync/dtchannel/dtchannel.go
```diff
@@ -262,6 +282,113 @@ func (c *Channel) Resume(ctx context.Context, extensions []graphsync.ExtensionDa
...
+
+	// set data limit
+	c.dataLimit = chst.DataLimit()
+	return nil
+}
+
+// ActionFromChannelState comparse internal graphsync channel state with the data transfer
+// state and determines what if any action should be taken on graphsync
+func (c *Channel) ActionFromChannelState(chst datatransfer.ChannelState) Action {
+	c.lk.Lock()
```
rvagg, line 358: just an `RLock()` would be OK here wouldn't it?

transport/graphsync/hooks.go
```diff
@@ -13,44 +16,24 @@ import (
 
 // gsOutgoingRequestHook is called when a graphsync request is made
 func (t *Transport) gsOutgoingRequestHook(p peer.ID, request graphsync.RequestData, hookActions graphsync.OutgoingRequestHookActions) {
-	message, _ := extension.GetTransferData(request, t.supportedExtensions)
 
-	// extension not found; probably not our request.
-	if message == nil {
+	chid, ok := t.requestIDToChannelID.load(request.ID())
```
rvagg, line 20: can you add (back) a comment here to say why we might not get one for this `request.ID()`? is it simply going to be a spoof (or mistake) from the peer?

transport/graphsync/initiating_test.go
```diff
@@ -0,0 +1,1325 @@
...
+	})
+
+	t.Run("cleanup request", func(t *testing.T) {
+		th.Transport.CleanupChannel(th.Channel.ChannelID())
+		require.Len(t, th.DtNet.UnprotectedPeers, 1)
+		require.Equal(t, th.DtNet.UnprotectedPeers[0], testharness.TaggedPeer{th.Channel.OtherPeer(), th.Channel.ChannelID().String()})
+	})
+}
+
+/*	"gs outgoing request with recognized dt push channel will record incoming blocks": {
```
rvagg, line 333: are there still pending tests in this comment block to be ported?

transport/graphsync/receiver.go
```diff
@@ -56,50 +59,39 @@ func (r *receiver) receiveRequest(ctx context.Context, initiator peer.ID, incomi
 		ch = r.transport.trackDTChannel(chid)
 	}
 
-	if receiveErr == datatransfer.ErrResume && ch.Paused() {
-
-		var extensions []graphsync.ExtensionData
+	if receiveErr != nil {
 		if response != nil {
```
rvagg, line 63: What's happening in here now if `response == nil`, is it right that we continue on (i.e. that this is really `if receiveErr != nil && response != nil {`. It looks like we can get all the way down to the end and do a `r.transport.processAction()` with a `nil` `response`.

rvagg (approved):
So, this is fine I guess; it's way too large with too much change to properly review without a more intimate understanding of the codebase so I've done my best to keep up but mostly I'm left commenting on isolated pieces rather than the architectural aspects of it. Overall it _seems_ good, there's some good simplification and pushing some concerns down into the transports seem like they're done well. The channel state migration looks good I think. There's some pieces of commented out code that makes it look incomplete (tests, and the "legacy" transport events which should probably be removed if we're not seeing a path to using them).

> Was it worth it?

_Feels_ like an incremental improvement from an architectural quality standpoint. But it's too complex for me to give a judgement beyond that. I hope you think it was worth it since you put in the effort!

impl/impl.go
hannahhoward: I hate the lack of terneries in go too. on this we agree!

transport.go
hannahhoward: so my original plan, which I'd love your feedback on, was to make the whole thing an event stream, which would remove the requirement of any synchronous hooks. As it turns out, that's kinda hard. This was an in between point when I thought may be I could make this all work for just incoming new or restart requests, which I may still consider. But, honestly, this was the first final interface I could get stood up and running after trying several times!

transport/graphsync/initiating_test.go
hannahhoward: no they should go. finishing those out right now.

transport/graphsync/hooks.go
hannahhoward: fair enough

hannahhoward:
A bit of background on the commented out events.

Where I would like to get to is no synchronous hooks. Just one OnTransportEvent function.

My baseline experience with these layers of transports is as follows: the more we expect/need to handle events synchronously, the more inflexible the code becomes, and the more likely we end up with race conditions. My idea for the future is one kinda like React in JS if you've worked with that -> data/parameters down + events up. So go-data-transfer should say to the transport: here's what I want to happen, and the transport should say: here's what actually did happen.

The problem is we just aren't quire there yet. I tried to figure this one out and just wasn't prepared to deal with the actuality that GraphSync's current hooks are synchronous. Or, at least, not in a single PR that was huge right?

hannahhoward:
> Was it worth it?
> Feels like an incremental improvement from an architectural quality standpoint. But it's too complex for me to give a judgement beyond that. I hope you think it was worth it since you put in the effort!

The hardest part here is I definitely can't keep banging away at this, but this is definitely not complete, and won't be complete till we implement a second transport. That's where the rubber will hit the road. Still, overall, I feel better about this. I just wish it weren't so hard. To some extent, it may be an argument that go-data-transfers goals are overly ambitious.

transport.go
```diff
@@ -17,75 +17,118 @@ const LegacyTransportID TransportID = "graphsync"
...
-	// OnReceiveDataError is called when a network error occurs receiving data
-	// at the transport layer
-	OnReceiveDataError(chid ChannelID, err error) error
+	// OnTransportEvent is dispatched when an event occurs on the transport
+	// It MAY be dispatched asynchronously by the transport to the time the
+	// event occurs
+	// However, the other handler functions may ONLY be called on the same channel
+	// after all events are dispatched. In other words, the transport MUST allow
+	// the handler to process all events before calling the other functions which
+	// have a synchronous return
```
rvagg, line 126: From your latest comments in the thread, I think this comment is getting at the point you made:

> Where I would like to get to is no synchronous hooks. Just one OnTransportEvent function.

Which points to the deficiency in the current half-way state? I take it that you've chewed off the doable events into `OnTransportEvent()`, with the remaining ones TODO.

So we still have:

* OnRequestReceived
* OnRequestReceived
* OnContextAugment

And you've experimented with splitting `OnRequestReceived` into `Update`, `Pause` and `Voucher` forms.

The context augment one being synchronous makes sense, the other ones do stand out a little now so I see the awkwardness of not going the full way on bundling everything into the single event _channel_. But, I think it's reasonable to stop where you have just to get this over the line.

However, it would be good to add some more inline comments about this state as a reminder for later or a hint for the next person picking this up about (a) the fact of the half-way state and (b) the ideal of moving toward fully async eventing.

impl/events.go
```diff
@@ -18,128 +14,77 @@ import (
 
 // OnChannelOpened is called when we send a request for data to the other
 // peer on the given channel ID
```
rvagg, line 16:
```suggestion
// OnTransportEvent is dispatched when an event occurs on the transport
```

transport.go
hannahhoward: absolutely. yes, the hard part about the others being asynchronous is that graphsync has synchronous hooks. so we'll need a bit of channel magic. messing with all this at once was just creating a back and forth with no progress being made.

PR #343, opened 2022-08-10
Fix close functions

hannahhoward:
# Goals

fix close functions post v2 upgrade

# Implementation

- Switch from checking status for transfer complete in transport as an indicator of a closed transfer to checking a direct function on channel state: TransferClosed()
- Have existing transfer closing events set the transfer to be closed
- Add an additional event, CloseTransfer, that dispatches to simply mark the transfer as closed without changing status
- Have the closing functions only call CloseTransfer before actually updating the transport, and then call the final termination functions afterward.

### DIFF
--- a/channels/channel_state.go
+++ b/channels/channel_state.go
@@ -158,6 +158,10 @@ func (c channelState) SelfPaused() bool {
 	return c.ResponderPaused()
 }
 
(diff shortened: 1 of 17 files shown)


hannahhoward:
This isn't going to work :(

PR #347, opened 2022-09-19
update to newer libp2p

willscott:
needs updated graphsync to be linked against


willscott:
the updated graphsync I pulled up to the current libp2p (https://github.com/ipfs/go-graphsync/pull/392) is only going to work against v2 of this library, namely #323 

what's the plausibility of switching over to the v2 branch as part of the libp2p upgrade? i'm having trouble understanding if there's a bunch of outstanding work.

I'd really prefer not trying to do this against as divergent branches from usable current versions of graphsync/datatransfer

rvagg:
> what's the plausibility of switching over to the v2 branch as part of the libp2p upgrade? i'm having trouble understanding if there's a bunch of outstanding work.

I think we're mostly done _except_ for integration and testing which is a bit of a big deal since it's all breaking changes. There were some additional features @hannahhoward wanted to try and get in here IIRC but mostly it's a matter of being able to justify the time to get it all working and pushed through the stack and justifying that time on top of everything else seems pretty tricky at the moment. There are some things that are coming up in retrieval failures that would be easier to deal with if we had this done, so maybe we can get slot it in to our priorities?

willscott:
I think at this point, recognizing that a v2 branch has been left unmerged since may, so 4 months, i'm going to push hard against adding any additional features before merge. We should merge an MVP and iterate so that the continued stream of things like this libp2p refactor have some possibility of getting trickled through rather than blocking on these epic part-way-finished refactors.

PR #350, opened 2022-10-07
Upgrade libp2p v0.22 onto v2

hannahhoward:
# Goals

upgrade v2 revised version to libp2p v0.22
also upgrades graphsync and removes go-libp2p-core paths


willscott (approved):
this looks largely mechanical. seems fine.

Issue #355, opened 2023-01-06
OpenDataTransferChannel should accept an event subscriber that only gets events for the channel


hannahhoward:
# Goals

- remove perpetual race condition around "how do I subscribe to events for just my channel"

# What

- OpenDataTransferChannel should take an optional call back for events
- Only events on the channel should be dispatched

# How

- implement a single global subscribe in go-data-transfer that dispatches to individual retrieval subscribers

PR #361, opened 2023-01-11
Subscriber per channel

hannahhoward:
# Goals

implement #355 

# Implementation

- Adds a channel subscriptions class to dispatch events to different channels
- Removes sending the Open event from CreateNew so subscriptions always catch all events on a channel
- Adds a TransferOptions var arg pattern to Open calls -- we can add WithTransportConfigurer soon!
- Adds an integration test
- bassed on #358 but with some more flexible impl

### DIFF
--- a/channelsubscriptions/channelsubscriptions.go
+++ b/channelsubscriptions/channelsubscriptions.go
@@ -1,0 +1,51 @@
+package channelsubscriptions
+
+import (
+	"sync"
+
+	datatransfer "github.com/filecoin-project/go-data-transfer/v2"
+	"github.com/filecoin-project/go-data-transfer/v2/channels"
+)
+
+type ChannelSubscriptions struct {
+	subscriptions   map[datatransfer.ChannelID][]datatransfer.Subscriber
+	subscriptionsLk sync.RWMutex
+	unsub           datatransfer.Unsubscribe
+}
+
+type SubscriptionAPI interface {
+	SubscribeToEvents(subscriber datatransfer.Subscriber) datatransfer.Unsubscribe
+}
+
+func NewChannelSubscriptions(subscriptionAPI SubscriptionAPI) *ChannelSubscriptions {
+	cs := &ChannelSubscriptions{
+		subscriptions: make(map[datatransfer.ChannelID][]datatransfer.Subscriber),
+	}
+	cs.unsub = subscriptionAPI.SubscribeToEvents(cs.subscriber)
+	return cs
+}
+
+func (cs *ChannelSubscriptions) Stop() {
+	cs.unsub()
+}
+
+func (cs *ChannelSubscriptions) Subscribe(chid datatransfer.ChannelID, cb datatransfer.Subscriber) {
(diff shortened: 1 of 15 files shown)


channelsubscriptions/channelsubscriptions.go
```diff
@@ -0,0 +1,51 @@
...
+	cs.unsub()
+}
+
+func (cs *ChannelSubscriptions) Subscribe(chid datatransfer.ChannelID, cb datatransfer.Subscriber) {
+	cs.subscriptionsLk.Lock()
+	defer cs.subscriptionsLk.Unlock()
+	cs.subscriptions[chid] = append(cs.subscriptions[chid], cb)
+}
+
+func (cs *ChannelSubscriptions) subscriber(evt datatransfer.Event, state datatransfer.ChannelState) {
```
rvagg, line 38: this feels a bit odd, could just be the name and the fact that it's for test only; was it meant to do more?

errors.go
```diff
@@ -30,3 +30,6 @@ const ErrRejected = errorType("response rejected")
 
 // ErrUnsupported indicates an operation is not supported by the transport protocol
 const ErrUnsupported = errorType("unsupported")
+
+// ErrAlreadySubscribed indicates a subscription to events exists for the given channel
+const ErrAlreadySubscribed = errorType("already subscribe to events for given channel id")
```
rvagg, line 35:
```suggestion
const ErrAlreadySubscribed = errorType("already subscribed to events for given channel id")
```

but this isn't used, was it meant to be or should it be deleted? looks like it'd be hard to check with the current API because we can't do equality checking.

impl/impl.go
```diff
@@ -189,6 +194,17 @@ func (m *manager) OpenPushDataChannel(ctx context.Context, requestTo peer.ID, vo
...
+
+	eventsCb := tc.EventsCb()
+	if eventsCb != nil {
+		m.channelSubscriptions.Subscribe(chid, eventsCb)
+	}
+
+	err = m.channels.Open(chid)
+	if err != nil {
+		return chid, err
+	}
```
rvagg, line 206: not necessary, just seeing an opportunity

```suggestion
	if eventsCb := tc.EventsCb(); eventsCb != nil {
		m.channelSubscriptions.Subscribe(chid, eventsCb)
	}

	if err := m.channels.Open(chid); err != nil {
		return chid, err
	}
```

impl/impl.go
```diff
@@ -217,19 +233,33 @@ func (m *manager) OpenPushDataChannel(ctx context.Context, requestTo peer.ID, vo
...
```
rvagg, line 261: ditto, take it or leave it

```suggestion
	if eventsCb := tc.EventsCb(); eventsCb != nil {
		m.channelSubscriptions.Subscribe(chid, eventsCb)
	}

	if err := m.channels.Open(chid); err != nil {
		return chid, err
	}
```

rvagg (approved):
doing something about `ErrAlreadySubscribed` would be good before merge

PR #362, opened 2023-01-16
Support per channel transport config

hannahhoward:
# Goals

This aims to import transport configuration by allowing a configuration when opening a channel. 

You can now pass WithTransportOptions, and then use options provided by your chose transport.

# Implementation

- track transport options in a simple mutex protected array
- **BREAKING** (but hey, it's go-data-transfer v2 which is still RC) - change the signature of RegisterTransportConfigurer, which was super weird, to return TransportOptions -- this reduces what the calling library has to do.
- Transports themselves now provide options which makes for MUCH cleaner usage.

# For Discussion

This implementation is... not quite perfect, but I think good enough for Lassie + go-fil-markets, the consumers.

The interaction between the old mechanism and the new mechanism is somewhat conflicting, and once I tried to replace the old mechanism completely, I ran into many of the perpetual design problems I struggle with in this library:
- how to support restarts after a reboot (i.e. from disk)
- how to support configure transfers for the receiver of an incoming request
- rather than go down yet another design hole like I did with the transport refactor that never got finished, I decided to call this good enough.

Companion PR for go-fil-markets: https://github.com/filecoin-project/go-fil-markets/pull/780

### DIFF
--- a/impl/environment.go
+++ b/impl/environment.go
@@ -25,4 +25,5 @@ func (ce *channelEnvironment) ID() peer.ID {
 func (ce *channelEnvironment) CleanupChannel(chid datatransfer.ChannelID) {
 	ce.m.transport.CleanupChannel(chid)
 	ce.m.spansIndex.EndChannelSpan(chid)
+	ce.m.transportOptions.ClearOptions(chid)
 }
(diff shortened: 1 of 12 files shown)


rvagg (approved):
👌 https://github.com/filecoin-project/lassie/pull/54 works nicely as far as I can tell

PR #366, opened 2023-02-11
Make states/events printable

hannahhoward:
# Goals

Resolves a long standing Quality Of Life improvement -- adds a fmt.Stringer methods to status and events so they appear in readable form when the state machine fails

I have no idea why I didn't do this earlier.

### DIFF
--- a/events.go
+++ b/events.go
@@ -1,6 +1,8 @@
 package datatransfer
 
-import "time"
+import (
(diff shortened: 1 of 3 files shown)


rvagg (approved):
huge improvement! I've been looking up and even counting to find the right events to see what's going on.

PR #367, opened 2023-02-11
Avoid sending bad transition

hannahhoward:
Fixed an issue handling finalization messages that were accidentally sending a pasue/resume event -- this was leading to an invalid state transition getting logged.


rvagg (approved):
So it's only the logging that gets impacted from this? It couldn't be the cause of some other weird state bugs we seem to have?

rvagg:
I guess it's complete by this point so there's not much other trouble it could cause?

hannahhoward:
yes that's correct. the state machine is still a mess, even though it's better in GTD2. I would like to just drop it eventually but we aren't there yet sadly :(

PR #371, opened 2023-03-09
fix: expose and use statemachine stop

rvagg:
only important if you want to fully cleanup after starting and stopping a datatransfer manager


hannahhoward (approved):
LGTM

PR #373, opened 2023-03-29
Update to the latest graphsync

gammazero:
Update graphsync to get the latest fixes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/ipfs/go-cid v0.3.2
 	github.com/ipfs/go-datastore v0.6.0
 	github.com/ipfs/go-ds-badger v0.3.0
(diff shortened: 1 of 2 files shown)


rvagg (dismissed):

rvagg:
those CI failures are pretty persistent across 3 CI re-runs on this, maybe not flakes so we either ought to figure that out, unless @hannahhoward knows whether they genuinely are flaky

rvagg:
Same failures experienced https://github.com/filecoin-project/go-data-transfer/pull/374 using an updated graphsync from https://github.com/ipfs/go-graphsync/pull/420 which is based on `main` there; the failures don't happen on the current `master` here, so they are new with the updated graphsync.

willscott (approved):

PR #374, opened 2023-04-04
feat: MaxLinks for pulls

rvagg:
Uses https://github.com/ipfs/go-graphsync/pull/420

Further comments in Lassie

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/ipfs/go-cid v0.4.1
 	github.com/ipfs/go-datastore v0.6.0
 	github.com/ipfs/go-ds-badger v0.3.0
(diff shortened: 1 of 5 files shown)


hannahhoward (commented):
My biggest concern is a transport option that only works one direction, with no indicator to that effect. I think the right thing is to update graphsync PR to go both directions, and then incorporate that here.

rvagg:
Failures discussed in https://github.com/filecoin-project/go-data-transfer/pull/373, they come from the diff between go-graphsync v0.14.0 and the current `main` there.

hannahhoward (commented):
LGTM

hannahhoward (approved):

PR #387, opened 2024-05-16
chore!: update graphsync, libp2p, quic; bump go version

rvagg:
Includes https://github.com/filecoin-project/go-data-transfer/pull/386


rvagg:
updated cbor-gen too, quite a few changes there, including map sorting and some other optimisations since it was last updated; the map sorting shouldn't be a problem though because it only impacts data that's never treated as an IPLD block so doesn't need a consistent hash

PR #389, opened 2025-10-15
upgrade go-graphsync and other dependencies

gammazero:
- replace test code with equivalents from ipfs/go-test


rvagg (approved):
Yeah, I guess. I'm not sure if anyone's actually consuming this any more. Boost forked this long ago to trim it down, maybe we're using it in Lassie still? fwiw this is probably the buggiest (and most complex) part of the stack and probably the reason why people unfairly blame graphsync for being bad. It would be good for it not to be used, we're not investing in this _at all_.
Need to fix the gen fix though.

gammazero:
Fixed the cbor gen stuff.

The problem is that IPNI still has to use go-data-transfer because there are still Filecoin data providers who have not yet updated to a newer boost or curio, and IPNI must still support indexing over data-transfer/graphsync.

@rvagg I do not have permission to merge.

